PR #188, opened 2019-11-05
Add navigation menu to UI

RamV13:
This adds a navbar that contains a page navigation hamburger menu.

This will be followed up by a change that includes a disaster picker in the top-left corner of the navbar.


RamV13:
Friendly ping for review

docs/navbar.html
```diff
@@ -0,0 +1,20 @@
...
+      <input id="nav-input" type="checkbox" />
+
+      <!-- Hamburger. -->
+      <span></span>
+      <span></span>
+      <span></span>
+
+      <ul class="nav-menu">
+        <a href="/"><li>Map</li></a>
+        <a href="/import/add_asset.html"><li>Upload Asset</li></a>
```
janakdr: Should probably link to add disaster page? I think this upload page might actually be less useful with some recent changes. Isn't it actually broken?

cypress/integration/integration_tests/nav_test.js
```diff
@@ -0,0 +1,14 @@
+describe('Integration test for the navbar', () => {
+  it('Open the nav menu and change pages', () => {
+    cy.visit(host);
+
+    // Click the burger.
+    cy.get('#nav-input').click();
+
+    // Click the add asset page link.
+    cy.get('#nav-menu a:first').next().click({force: true});
```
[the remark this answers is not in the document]
RamV13: For some reason, Cypress believes the element is not visible. I've also added a comment as well.

docs/navbar.html
RamV13: Switched

cypress/integration/integration_tests/nav_test.js
```diff
@@ -0,0 +1,15 @@
+describe('Integration test for the navbar', () => {
+  it('Open the nav menu and change pages', () => {
+    cy.visit('');
+
+    // Click the burger.
+    cy.get('#nav-input').click();
+
+    // Click the add asset page link. This is forced because Cypress mistakenly
```
janakdr: nit: now add disaster?

cypress/integration/integration_tests/nav_test.js
```diff
@@ -0,0 +1,15 @@
...
+    // believes the element is not visible.
```
janakdr: does it change if you do cy.awaitLoad()? Just to understand why.

docs/import/add_asset.js
```diff
@@ -425,3 +425,9 @@ function addFileToDelete(file) {
     filesToDelete.length = 0;
   }
 }
+
+$(() => {
```
janakdr: nit: I think this and the next lambda can both be one-lined :) Here and below. Maybe also extract to a function?

docs/navbar.html
```diff
@@ -0,0 +1,20 @@
+<div class="navbar">
+  <h1 id="nav-header" class="nav-header"></h1>
+
+  <nav role="navigation">
+    <div id="nav-toggle">
+      <!-- A hidden checkbox just to be the click receiver the :checked selector can be used. -->
```
janakdr: nit: not entirely grammatical

cypress/integration/integration_tests/nav_test.js
RamV13: No it doesn't unfortunately. It's because the `<a>` tag's dimensions are 0x0. I guess I could fiddle with the CSS more but I'm not sure if it's worth it.

docs/navbar.html
RamV13: haha thanks

docs/navbar.js
```diff
@@ -0,0 +1,10 @@
+export {loadNavbar};
+
+/**
+ * Loads the navbar and calls the callback upon load.
+ *
+ * @param {Function} callback the function called upon load
+ */
+function loadNavbar(callback) {
```
janakdr: it looks like the callbacks for this function are all of the same form. Could we just pass the string in as an argument and have this construct the callback?

docs/import/add_disaster.js
```diff
@@ -327,3 +328,6 @@ function createOptionFrom(innerTextAndValue) {
       .val(innerTextAndValue)
       .prop('id', innerTextAndValue);
 }
+
+loadNavbar(
+    () => document.getElementById('nav-header').innerHTML = 'Add Disaster');
```
janakdr: is there a reason not to use jquery syntax here?

cypress/integration/integration_tests/nav_test.js
janakdr: Definitely not worth it :) just wanted to understand the cause! Maybe add a few words to the comment saying this about the <a> tag?

docs/navbar.js
RamV13: In #218, the child PR, this is not the case that's why I set it up like this. Fixed in this PR.

janakdr:
Unfortunately this broke add_disaster_test, because this PR makes $ a global variable that needs to be loaded at script startup. Trying to see if there's an easy fix.

janakdr:
Mailed https://github.com/givedirectly/Google-Partnership/pull/225

PR #190, opened 2019-11-06
Restructure Firestore data to avoid document with unknown collections

janakdr:
As we've discussed offline, the Firebase web client cannot list the collections in a document, which means that, since in our current schema a year is a document, we can't list the disasters from that year. To get around this, make the structure disaster-metadata/<year>-<disaster> and then put all disaster metadata (except for usershapes, which remain global, hmm...) into fields in the <year>-<disaster> document.

I used a script for this that I'd rather not clean up for submission, since this should be a one-off. But if we need to do this again, I can push it up.

Also fix a sequential processing issue: we were blocking score computation, the most expensive part of our page load, on Firestore retrieval, because we wanted to know how many layers there were. But since score always goes last, we can just do a bit of special-casing in layer_util.js to avoid needing to know how many layers there are ahead of time.

I tested that we're actually using the new data by changing a layer to be true by default in the new data and verifying that it was really there.

### DIFF
--- a/docs/run.js
+++ b/docs/run.js
@@ -203,7 +203,10 @@ function addLayers(map) {
  */
 function addScoreLayer(layer) {
   addLayerFromGeoJsonPromise(layer, scoreLayerName, 'lastElement');
-  // Checkbox may not exist yet if layer metadata not retrieved yet.
+  // Checkbox may not exist yet if layer metadata not retrieved yet. The
+  // checkbox creation will check the box by default. We check it here in case
+  // it was unchecked by the user, and this is coming from a weight/threshold
+  // update.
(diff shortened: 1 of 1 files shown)


juliexxia:
Sorry, my bad for not catching this but I think this PR also need to change https://github.com/givedirectly/Google-Partnership/blob/1d5a28e6dd77eedd1a2c4a4717bb8e1623193545/docs/import/center.js#L48

We're probably changing up the import script quite a bit but I think this will come with it

janakdr:
My bad! https://github.com/givedirectly/Google-Partnership/pull/198 fixes the center.js issue.

Issue #171, opened 2019-10-29
Allow users to toggle between disasters


ruthtalbot:
- Allow users to switch to a new disaster and load that new disasters data
- Allow users to set a default disaster

janakdr:
I think only remaining issue here is the discrepancy between the way manage_layers does this and the way the rest of the system does it.

janakdr:
Closed by #299

Issue #170, opened 2019-10-29
Add 'Add Disaster' page


ruthtalbot:
Landing page to allow users to add a new disaster, including name, type, location, and assets(?).

PR #195, opened 2019-11-08
[1/n] Begin add_disaster.html

juliexxia:
First PR for the process of adding a new disaster. This takes care of the drop down menu that reads in our disasters from firestore as well as the form to add to that drop down menu. Upon creating a new disaster or selecting an existing one, we also surface pickers for each state associated with the disaster. In future PRs will include functionality to pick from these assets to add as data or layers to the disaster. 

#170, #171


cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -0,0 +1,152 @@
...
+          expect(picker).to.contain(
+              AddDisaster.gdEePathPrefix + 'states/' + KNOWN_FAKE_STATE +
+              '/snap');
+          expect(picker.children().length).to.equal(2);
+          expect($('#' + UNKNOWN_FAKE_STATE + '-adder').children().length)
+              .to.equal(1);
+        });
+  });
+
+  it.only('adds a new disaster to firestore', () => {
```
juliexxia: test in q

docs/import/add_disaster.js
```diff
@@ -0,0 +1,232 @@
...
+  }
+}
+
+/**
+ *
+ * @param disasterId
+ * @param states
+ * @return {Promise<void>}
+ */
+function writeDisaster(disasterId, states) {
```
juliexxia: code in q

juliexxia: code in q

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/**
+ * Writes the given details to a new disaster entry in firestore. Fails if
+ * there is an existing disaster with the same details.
+ * @param {string} disasterId of the form <year>-<name>
+ * @param {Array<string>} states array of state (abbreviations)
+ * @return {Promise<boolean>} returns true after successful write to firestore.
+ */
+function writeDisaster(disasterId, states) {
```
janakdr: nit: maybe writeNewDisaster or addDisaster to disambiguate from modifying an existing disaster?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+  if (disasters.has(disasterId)) {
```
janakdr: do you want to do some text validation of disasterId? It should at least be a valid Firestore doc name

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+    return Promise.resolve(false);
+  }
+
+  disasters.set(disasterId, states);
+  toggleState(disasterId);
+
+  clearStatus();
+  const disasterOptions = $('#disaster > option');
+  // comment needed to quiet eslint on no-invalid-this rules
+  disasterOptions.each(/* @this HTMLElement */ function() {
```
janakdr: might be overkill, but since this is a sorted list, binary search is going to be more efficient than linear iteration

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+// pickers.
+/**
+ * Requests all assets in ee directories corresponding to given states and
+ * displays them in pickers. Right now, selecting on those pickers doesn't
+ * actually do anything.
+ * @param {Array<string>} states e.g. ['WA']
+ * @return {Promise<void>} completes when we've finished filling all state
+ * pickers.
+ */
+function createStateAssetPickers(states) {
```
janakdr: Can we rephrase this function into two parts? One does the ee.listAssets and stores the data in the result of the Promise. Then another function takes the Promise and puts the data into the page. That way, the result of the Promise can be cached, removing the performance issue of listing assets when you're switching back and forth, and the two functions can be tested independently.

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+  const assetPickerDiv = $('#asset-pickers');
+  assetPickerDiv.empty();
+  return ee.data.listAssets(eeLegacyPathPrefix + 'states', {}, emptyCallback)
+      .then((result) => {
+        const folders = new Set();
+        for (const folder of result.assets) {
+          folders.add(folder.id.substring((gdEePathPrefix + 'states/').length));
```
janakdr: nit: extract this number to a constant? Looks like it can be global?

docs/import/add_disaster_startup.js
```diff
@@ -0,0 +1,21 @@
...
+const taskAccumulator = new TaskAccumulator(2, enableWhenReady);
+
+// TODO: do something with this promise, pass it somewhere - probably once
+// tagging happens.
+const firebaseAuthPromise = new SettablePromise();
+const authenticator = new Authenticator(
+    (token) => firebaseAuthPromise.setPromise(authenticateToFirebase(token)),
+    () => {
+      // Necessary for listAssets.
+      ee.data.setCloudApiEnabled(true);
```
janakdr: nit: can't this be at the top level?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -0,0 +1,162 @@
...
+        })
+        .then((success) => {
+          expect(success).to.be.false;
+          expect(status.is(':visible')).to.be.true;
+          expect(status.text()).to.eql('Error: Year must be a number.');
+
+          year.val('2000');
+          return addDisaster();
+        })
+        .then((success) => {
```
janakdr: nit: remove braces from one-liner

docs/import/add_disaster.js
juliexxia: Nice done.

docs/import/add_disaster.js
juliexxia: which number? the (gdEePathPrefix + 'states/').length) ? I think this is the only place it's used.

docs/import/add_disaster.js
```diff
@@ -0,0 +1,220 @@
...
+// Currently a map of disaster name to states. This pulls once on firebase
+// authentication and then makes local updates afterwards so we don't need to
+// wait on firebase writes to read new info.
+const disasters = new Map();
+
+const SENTINEL_OPTION_VALUE = '...';
+const SENTINEL_NEW_DISASTER_VALUE = 'NEW DISASTER';
+
+/**
+ * Populates the disaster picker with disasters from firestore, and enables
```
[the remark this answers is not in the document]
janakdr: I think you can set the size of the disaster picker so that it doesn't jump around. We don't know how long disaster names will be, but can make a pretty good guess?

docs/import/add_disaster.html
```diff
@@ -0,0 +1,97 @@
...
+      <option value="TX">Texas</option>
+      <option value="UT">Utah</option>
+      <option value="VT">Vermont</option>
+      <option value="VA">Virginia</option>
+      <option value="WA">Washington</option>
+      <option value="WV">West Virginia</option>
+      <option value="WI">Wisconsin</option>
+      <option value="WY">Wyoming</option>
+    </select><br>
+<!--    <button type="button" id="add-disaster-button" disabled="true">Add Disaster</button>-->
```
janakdr: remove?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+ * the ability to add a new disaster.
+ * @return {Promise<firebase.firestore.QuerySnapshot>}
+ */
+function enableWhenReady() {
+  // enable (currently hidden) add disaster button now that firestore is ready.
```
janakdr: remove? And rename method?

docs/import/add_disaster.js
janakdr: This function is called multiple times, so might as well extract to a constant.

docs/import/add_disaster_startup.js
```diff
@@ -0,0 +1,23 @@
...
+
+export {taskAccumulator};
+
+// 2 tasks: EE authentication, page load
+const taskAccumulator = new TaskAccumulator(2, enableWhenReady);
+
+// Necessary for listAssets.
+ee.data.setCloudApiEnabled(true);
+
+// TODO: do something with this promise, pass it somewhere - probably once
```
janakdr: Isn't this promise a prerequisite for anything else with Firebase happening? Shouldn't enableWhenReady not run until this happens?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+  addDisasterButton.on('click', addDisaster);
+
+  // populate disaster picker.
+  return getFirestoreRoot()
+      .collection('disaster-metadata')
+      .get()
+      .then((querySnapshot) => {
+        const disasterPicker = $('#disaster');
+        disasterPicker.append(createOptionFrom(SENTINEL_OPTION_VALUE));
+        querySnapshot.forEach((doc) => {
```
janakdr: Can we sort this in reverse order so most recent is at the top? Then we can also leave off the placeholder, which should simplify the code

docs/import/add_disaster.js
janakdr: Discussed offline, I think lowercase a-z seems like a good restriction, in the function below

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+
+// Needed for testing :/
+const emptyCallback = () => {};
+
+// TODO: add functionality for adding assets to disaster records from these
+// pickers.
+/**
+ * Requests all assets in ee directories corresponding to given states.
+ * @param {Array<string>} states e.g. ['WA']
+ * @return {Promise<void>} completes when we've finished filling all state
```
janakdr: Is this doc still up to date?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+      .then((result) => {
+        const folders = new Set();
+        for (const folder of result.assets) {
+          folders.add(folder.id.substring((gdEePathPrefix + 'states/').length));
+        }
+        const promises = [];
+        for (const state of states) {
+          const dir = eeLegacyPathPrefix + 'states/' + state;
+          if (!folders.has(state)) {
+            ee.data.createFolder(dir, false, emptyCallback);
```
janakdr: can we tell when this create call finishes?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+            promises.push(Promise.resolve([state, []]));
+          } else {
+            promises.push(
+                ee.data.listAssets(dir, {}, emptyCallback).then((result) => {
+                  const assets = [];
+                  if (result.assets) {
+                    for (const asset of result.assets) {
```
janakdr: do folders have ids? Do we need to filter them out, in case someone creates a subfolder? Should we display a warning to the user that subfolders won't be recursed into?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,253 @@
...
+  } else {
+    const states = disasters.get(disaster);
+    const statesToFetch = [];
+    for (const state of states) {
+      if (!stateAssets.has(state)) statesToFetch.push(state);
+    }
+
+    // TODO: add functionality to re-pull all cached states from ee without
+    // reloading the page.
+    const assetPickersDone = getAssetsFromEe(statesToFetch).then((assets) => {
```
janakdr: should probably avoid this call if statesToFetch is empty

cypress/integration/unit_tests/add_disaster_test.js
janakdr: Don't see it?

docs/import/add_disaster.js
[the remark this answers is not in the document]
juliexxia: Set at 160 px for now - not very responsive but I'm ok with that as a first go

docs/import/add_disaster.js
juliexxia: Sorry - did mean to leave as disabled and enable here.

docs/import/add_disaster.js
juliexxia: Done. Most recent is automatically displayed when you open the page now.

docs/import/add_disaster.js
juliexxia: sgtm

docs/import/add_disaster.js

docs/import/add_disaster.js
juliexxia: Updated.

docs/import/add_disaster.js

docs/import/add_disaster.js
juliexxia: Added a silent check that it's an image, imagecollection, or table. I'm not too too fussed about warning the user it won't load subdirectories because I think that's sort of clear from the basic nature of the asset picker that that wouldn't work. I also worry about a warning being spammy.

\+ added a test

docs/import/add_disaster_startup.js
juliexxia: Yes......tbh not sure what I was thinking here. Thanks.

docs/import/add_disaster.js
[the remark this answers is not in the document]
janakdr: I'm still seeing it very small, at least in my browser that is throwing an authentication error. You can see it small if you open this page in an incognito window.

image

docs/import/add_disaster.js
```diff
@@ -0,0 +1,289 @@
...
+        }
+        createAssetPickers(states);
+      });
+    }
+
+    // TODO: display more disaster info including current layers etc.
+    $('#new-disaster').hide();
+    $('#selected-disaster').show();
+    const assetPickerDiv = $('#asset-pickers');
+    // We are already doing this inside createAssetPickers but not until
```
janakdr: nit: move this up before the statesToFetch.length check and you can avoid doing this unless you're in the else block?

docs/import/add_disaster_startup.js
janakdr: I don't think you'll need to pass it anywhere, so you could just scrap the SettablePromise and make the token receiver:

(token) => authenticateToFirebase(token).then(() => taskAccumulator.taskCompleted())

docs/import/add_disaster.js
```diff
@@ -0,0 +1,289 @@
...
+
+// Needed for testing :/
+const emptyCallback = () => {};
+
+// TODO: add functionality for adding assets to disaster records from these
+// pickers.
+/**
+ * Requests all assets in ee directories corresponding to given states.
+ * @param {Array<string>} states e.g. ['WA']
+ * @return {Promise<Array<string | Array<string>>>} 2-d array of all retrieved
```
janakdr: nit: I think this signature should be Promise<Array<Array<string, Array<string>>>>, right? Because it's an array (one element per state), and each element of that array is itself a 2-element array, first element state, and second element array of strings.

docs/import/add_disaster.js
```diff
@@ -0,0 +1,289 @@
...
+
+// Currently a map of disaster name to states. This pulls once on firebase
+// authentication and then makes local updates afterwards so we don't need to
+// wait on firebase writes to read new info.
+const disasters = new Map();
+
+// Map of state to list of known assets
+const stateAssets = new Map();
+
+const SENTINEL_OPTION_VALUE = '...';
```
janakdr: Do we still need this?

docs/import/add_disaster.html
```diff
@@ -0,0 +1,99 @@
...
+  <span> </span>
+  <button type="button" id="create-new-disaster">Create New Disaster</button>
+</form><p></p>
+<div id="status"></div>
+<div id="new-disaster" hidden="true">
+  <form action="" id="new-disaster-form">
+    <label for="name">Disaster Name: </label><input id="name"><label for="name"><i> ex: harvey</i></label><br>
+    <label for="year">Disaster Year: </label><input id="year"><label for="year"><i> ex: 2017</i></label><br>
+    <!-- TODO(#207): support us territory data sets and add them here -->
+    <label for="states">State(s): </label><select multiple="multiple" id="states">
```
janakdr: This picker is pretty small. Maybe make it a bit bigger by default?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,289 @@
...
+ * @param {Array<string>} states e.g. ['WA']
+ */
+function createAssetPickers(states) {
+  const assetPickerDiv = $('#asset-pickers');
+  for (const state of states) {
+    const assetPicker = $(document.createElement('select')).attr({
+      multiple: 'multiple',
+      id: state + '-adder',
+    });
+    assetPicker.append(createOption(
```
janakdr: Why should this be an option, as opposed to just some text on top of the box?

docs/import/add_disaster.js
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Given states, displays their assets in pickers. Right now, selecting on
+ * those pickers doesn't actually do anything.
+ * @param {Array<string>} states e.g. ['WA']
+ */
+function createAssetPickers(states) {
+  const assetPickerDiv = $('#asset-pickers');
+  for (const state of states) {
+    const assetPicker = $(document.createElement('select')).attr({
+      multiple: 'multiple',
```
janakdr: I know this is going to change a lot, but can you tell me the use case for picking multiple assets?

juliexxia: I think this select box is just for adding assets - I image below this on the page there will be all the assets you've selected as relevant for the current disaster. This way you can mark multiple assets as relevant at a time. This select box won't be for any actually tagging/editing of how to use the elements.

docs/import/add_disaster.js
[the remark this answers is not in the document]
juliexxia: I actually don't see it as small from the GD account, or from an incognito page, or my corp page. It's always 160 px.

docs/import/add_disaster.js
juliexxia: I was using it for the upload additional assets option. Now we don't anymore.

docs/import/add_disaster.js
juliexxia: We can give it a alert a butter bar or something in the callback. But I'd prefer for the user not to have to wait for folder creation to see the empty select. I've also added the set acls call so that's where the status update can go too.

docs/import/add_disaster.js
```diff
@@ -194,7 +192,12 @@ function getAssetsFromEe(states) {
         for (const state of states) {
           const dir = legacyStateDir + '/' + state;
           if (!folders.has(state)) {
+            // This will print a console error for anyone other than the gd
+            // account. Ee console seems to have the power to grant write access
+            // to non-owners but it doesn't seem to work. Sent an email to
+            // gestalt.
             ee.data.createFolder(dir, false, () => {
+              // TODO: add status bar for when this is finished.
               ee.data.setAssetAcl(dir, {all_users_can_read: true});
```
janakdr: Looks like this is deprecated, and setIamPolicy is the new hotness

docs/import/add_disaster.js
janakdr: So these are assets that are not actually part of this disaster yet?

docs/import/add_disaster.html
janakdr: I meant higher, not wider :)

docs/import/add_disaster.js
```diff
@@ -0,0 +1,292 @@
...
+    $('#selected-disaster').hide();
+  } else {
+    const states = disasters.get(disaster);
+    const statesToFetch = [];
+    for (const state of states) {
+      if (!stateAssets.has(state)) statesToFetch.push(state);
+    }
+    // TODO: add functionality to re-pull all cached states from ee without
+    // reloading the page.
+    let assetPickersDone = Promise.resolve();
```
janakdr: Switching to a disaster should also pull from Firestore, right? Can you leave a TODO or skeleton of how/where that would happen? I think it will complicate this logic enough that it might be worth taking that into account here.

docs/import/add_disaster.js
juliexxia: So i get a pretty gross error when trying to use setIamPolicy. I actually get this also with the gd account and pretty unclear to me what it's about. So I'm going to stick with setAclAsset for now, leave a TODO and update the email to gestalt. 

image

docs/import/add_disaster.js
juliexxia: My basic idea is we keep the entire disaster-metadata document for each disaster in the disasters map values the way we're keeping the states right now. That way, we never have to read from firestore except for the first time in the beginning so we don't have to wait and make sure writes are done. I don't think it'll be too too much data and never having to do reads sounds pretty nice. In that case, I don't think there's too much to do here right now. Thoughts?

docs/import/add_disaster.html
juliexxia: Ahh, done.

docs/import/add_disaster.js
juliexxia: correct - ideally once you make an asset part of a disaster it gets taken off that list. Currently these pickers are entirely state-based and don't have any affiliation with the actual disaster (yet)

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -0,0 +1,184 @@
...
+    disasters.set('2003-spring', []);
+  });
+
+  afterEach(() => {
+    stateAssets.clear();
+    disasters.clear();
+  });
+
+  it('gets state asset info from ee', () => {
+    getAssetsFromEe([KNOWN_STATE, UNKNOWN_STATE]).then((assets) => {
```
janakdr: can you cy.wrap() this first promise (you can leave the then() unchanged). If this promise never resolved, the test would actually pass without that. Here and below.

docs/import/add_disaster.js
```diff
@@ -0,0 +1,299 @@
...
+
+// Currently a map of disaster name to states. This pulls once on firebase
+// authentication and then makes local updates afterwards so we don't need to
+// wait on firebase writes to read new info.
+const disasters = new Map();
+
+// Map of state to list of known assets
+const stateAssets = new Map();
+
+const SENTINEL_NEW_DISASTER_VALUE = 'NEW DISASTER';
```
janakdr: as discussed offline, I think you can remove this code.

docs/import/add_disaster.html
```diff
@@ -0,0 +1,100 @@
...
+  <script src="https://www.gstatic.com/firebasejs/7.2.1/firebase-auth.js"></script>
+  <!-- Load jQuery, a 3rd-party library. -->
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
+  <script src="https://apis.google.com/js/api.js"></script>
+  <script type="module" src="./add_disaster_startup.js"></script>
+</head>
+<body>
+
+<form action="" id="disaster-picker">
+  <label for="disaster">Disaster: </label>
```
janakdr: it might look a bit better to have the current disasters and "Create New Disaster" on separate lines. It also might be nice to gray out or otherwise indicate that the apparently currently selected disaster isn't actually relevant during disaster creation. Maybe making it invisible and then having a "cancel" button in the form. Right now, if you happen to click on the drop-down disaster list you'll lose any work you might have in progress on disaster creation. Maybe add a TODO?

juliexxia: I actually like having create new disaster on the same line, I did that on purpose because i think it's nice and tidy up there. I don't think it should be below (covered by the drop down) and above also seems slightly weird because I think the picker should be front and center. 

Agree though that it's confusing to have any old disaster in the picker while creating a new one. Added a todo to add_disaster_startup. to fix in a follow up.

PR #197, opened 2019-11-08
Increase page loading timeout for tests

janakdr:
Seeing some flaky failures on Travis, hoping this fixes them.


janakdr:
Upped to 200 seconds. I think there's something else happening, but I've seen tests taking >140 seconds, so this should help in some cases anyway.

PR #198, opened 2019-11-08
Fix center.js to use updated Firestore location for map bounds

janakdr:
Also let user know that map bounds calculation is in progress and when it completes. And add a unit test for center.js (first test of import functionality!).

First work on #175

### DIFF
--- a/docs/import/center.js
+++ b/docs/import/center.js
@@ -39,7 +39,7 @@ function storeCenter(features, firebaseAuthPromise) {
 /**
  * Writes the calculated bounds to firestore.
  * @param {Array<number>} bounds
- * @return {Promise<Array{number}>} Promise that completes with array of bounds
+ * @return {Promise<Array<number>>} Promise that completes with array of bounds
  * after Firestore write is complete.
  */
 function saveBounds(bounds) {
@@ -47,7 +47,7 @@ function saveBounds(bounds) {
     'map-bounds': {
       sw: new firebase.firestore.GeoPoint(bounds[0], bounds[1]),
       ne: new firebase.firestore.GeoPoint(bounds[2], bounds[3]),
-    }
+    },
   };
   return disasterDocumentReference()
       .set(docData, {merge: true})
--- a/docs/import/center.js
+++ b/docs/import/center.js
@@ -19,7 +19,7 @@ function withGeo(feature) {
  * @param {ee.FeatureCollection} features the featureCollection around which you
  * want to orient the map
  * @param {Promise} firebaseAuthPromise
- * @return {Promise<Array{number}>} Promise that completes with array of bounds
+ * @return {Promise<Array<number>>} Promise that completes with array of bounds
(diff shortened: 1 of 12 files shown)


docs/import/center.js
```diff
@@ -20,37 +19,37 @@ function withGeo(feature) {
...
     damageWithCoords.aggregate_min('lng'),
+    damageWithCoords.aggregate_max('lat'),
+    damageWithCoords.aggregate_max('lng'),
   ]);
-  Promise.all([firebaseAuthPromise, convertEeObjectToPromise(outerBounds)])
-      .then((results) => saveBounds(results[1]))
-      .catch(createError('error finding bounds of map'));
+  return Promise
+      .all([firebaseAuthPromise, convertEeObjectToPromise(outerBounds)])
+      .then((results) => saveBounds(results[1]));
```
juliexxia: Why are we not catching errors anymore?

docs/import/center.js
```diff
@@ -20,37 +19,37 @@ function withGeo(feature) {
...
  */
 function storeCenter(features, firebaseAuthPromise) {
   const damageWithCoords = ee.FeatureCollection(features.map(withGeo));
   // This is assuming we're not crossing the international date line...
   const outerBounds = ee.List([
-    damageWithCoords.aggregate_max('lat'),
-    damageWithCoords.aggregate_max('lng'),
     damageWithCoords.aggregate_min('lat'),
     damageWithCoords.aggregate_min('lng'),
+    damageWithCoords.aggregate_max('lat'),
```
juliexxia: why flipped?

docs/import/import_data.js
```diff
@@ -168,7 +168,19 @@ function attachBlockGroups(building, blockGroups) {
...
-  storeCenter(damage, firebaseAuthPromise);
+  const centerStatusLabel = document.createElement('span');
+  centerStatusLabel.innerText = 'Computing and storing bounds of map: ';
+  const centerStatusSpan = document.createElement('span');
+  centerStatusSpan.innerText = 'in progress';
+  storeCenter(damage, firebaseAuthPromise)
+      .then(trimGeoNumbers)
+      .then(
+          (bounds) => centerStatusSpan.innerText = 'Found bounds (' +
+              bounds[0] + ', ' + bounds[1] + '), (' + bounds[2] + ', ' +
```
juliexxia: Would be very cool to display the bounded map. A thought for later.

docs/import/center.js
janakdr: Because I think people generally process bounds as (min, max), and especially for LTR-native speakers, the left-most bound should come first. So, lower-left-to-upper-right is more natural than the reverse. Switched because this array is being displayed to users now.

docs/import/center.js
janakdr: Caught errors wouldn't propagate up and be visible to user

PR #199, opened 2019-11-11
Fix deletion of test data from Firestore

janakdr:
If the baseUrl is not set, then Cypress runs beforeEach hooks twice (https://github.com/cypress-io/cypress/issues/2636), which left lingering data.

With that fix, the belt-and-suspenders approach is mostly unnecessary, but added it for good measure: make the test directory start with the start date of the test and at the start of each test, delete all directories older than 24 hours. Unfortunately, Firestore has a nasty quirk: a document will only show up in its parent collection's listing if it has fields of its own, transitive fields don't count. So we add a dummy field to the root test document that we create, so that they show up in listings later.

This last part is obviously pretty hacky, and perhaps we don't actually need it with the root bug fixed. But I do worry about GiveDirectly accidentally going over Firestore quota if there's some bug in the future that makes tests fail without clean-up. If you think it's too hacky/confusing, though, I can revert it.

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -126,7 +126,7 @@ module.exports = (on, config) => {
       return Promise.all([harveyDoc.get(), signinPromise, deletePromise])
           .then((result) => Promise.all([
             documentReference.set(result[0].data(), {merge: true}),
-            documentReference.set({dummy: true}, {merge: true})
+            documentReference.set({dummy: true}, {merge: true}),
           ]))
           .then(
               () => Promise.all(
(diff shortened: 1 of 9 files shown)


cypress/plugins/index.js
```diff
@@ -178,6 +185,36 @@ function deleteTestData(currentTestRoot, app) {
...
+function deleteAllOldTestData(app) {
+  const currentDate = new Date();
+  const querySnapshotPromise = app.firestore().collection(testPrefix).get();
+  return querySnapshotPromise.then((queryResult) => {
+    const promises = [];
+    queryResult.forEach((doc) => {
+      const ref = doc.ref;
+      const testRunName = ref.id;
+      const dateElement = testRunName.split('-')[0];
+      const date = new Date(parseInt(dateElement, 10));
```
juliexxia: is the 10 necessary?

janakdr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

PR #200, opened 2019-11-11
Deflake test by making it a unit test

janakdr:
Most of these tests don't need to be integration tests. Starting with this one, since it's pretty flaky.

Setting an attribute on the event in overlaycomplete is a bit hacky. I can deduce when the promise is finished other ways as well, but since that promise is exactly what we want, I think it's nice to use it. Other alternatives would be to wait on the firebase add to complete (we can track that, since we're stubbing out the add), or to watch for the loading waiter changes (right now we completely stub those out).

The interaction between Cypress and Promises isn't my favorite, but not too terrible.

### DIFF
--- a/cypress/integration/unit_tests/polygon_draw_test.js
+++ b/cypress/integration/unit_tests/polygon_draw_test.js
@@ -183,7 +183,7 @@ describe('Unit test for ShapeData', () => {
     const map = new google.maps.Map(div, {center: {lat: 0, lng: 0}, zoom: 1});
     const polygon = new google.maps.Polygon({
       map: map,
-      paths: [{lat: 0, lng: 0}, {lat: 1, lng: 1}, {lat: 0, lng: 1}]
+      paths: [{lat: 0, lng: 0}, {lat: 1, lng: 1}, {lat: 0, lng: 1}],
     });
     setUpPopup();
     const popup = createPopup(polygon, map, 'notes');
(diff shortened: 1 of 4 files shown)


cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -176,6 +177,38 @@ describe('Unit test for ShapeData', () => {
     });
   });
 
+  it('Shows calculating before update finishes', () => {
+    const div = document.createElement('div');
+    document.getElementsByTagName('body')[0].appendChild(div);
```
juliexxia: Does document.body.appendChild work here? It's what we're using in update_test and might be nice to keep consistent.

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -176,6 +177,38 @@ describe('Unit test for ShapeData', () => {
...
+    firebaseCollection.add = () => Promise.resolve({id: 'id'});
+    const updatePromise = new SettablePromise();
+    cy.wrap(setUpPolygonDrawing(map, Promise.resolve()))
+        .then((drawingManager) => {
+          google.maps.event.trigger(drawingManager, 'overlaycomplete', event);
+          updatePromise.setPromise(event.resultPromise);
+        });
+    let calculatingDiv;
+    // Popup initialization happens async, apparently, and takes a bit of time.
+    cy.wait(50).then(() => {
```
juliexxia: Sorry, I'm not quite understanding - what is ensuring that these expects run before the calculating finishes?

janakdr: Added a comment.

PR #201, opened 2019-11-11
Set project license to MIT

janakdr:
Please let me know if you'd like a different license!


janakdr:
@mbarshay told me over Slack that this is ok :) Mind rubber-stamping?

PR #204, opened 2019-11-12
Special-case score rather than relying on it not being in firebaseLayers

janakdr:
In the case that the Firebase fetch is slow, firebaseLayers may not be initialized, but that shouldn't block the score from being displayed. I don't know if this is responsible for our flaky loading-forever issues on Travis, but it seems possible.


juliexxia:
This is nice because it's a more explicit check of what we're really thinking anyway.

PR #205, opened 2019-11-12
Switch to using EarthEngine hosted remotely

janakdr:
The proximate reason for this is that our unit tests are flakily failing (with a mysterious "SyntaxError: unexpected <") and loading the script from a remote origin rather than our local seems to fix it. But also EE people say that we should be using the "live" version of the library, rather than a downloaded version, and this seems like a step in the right direction.

Because we can now run all the tests on Chromium, and we have enough machines to make it worthwhile, switch back to parallel tests.

Also upgrade Cypress.


janakdr:
Actually, keeping tests as is, there's something weird about some of the new tests on polygon_draw_test on Chromium that I need to look into further.

PR #209, opened 2019-11-14
Stop using asset names as keys for layer metadata

janakdr:
Using asset names like this has a few disadvantages:
1. Not all assets have obvious names (map tiles, for instance).
2. We might want to treat the same FeatureCollection as two different layers in the future (say, pull a different property from it for each layer).
3. A fundamental property we want, that assets have unique indices, is not enforced by design in our data structure.
4. Having a global map (needed for looking up asset names) is a code smell.

Instead, using indices everywhere we can. This lets us get rid of the global firebaseLayers object entirely, which is nice. Also this spurs some clean-ups/restructurings in layer_util.js.

We need to pass the actual Firestore data object into toggleLayerOn, which makes me a little sad, but it's ok. (I think in the future we might want to render all layers eagerly, even if we don't display them: I think that Julie may have suggested this a while ago, and I pushed back? But I'm coming around, assuming we can handle the stress it would put on our local page.)

I've populated Firestore with the new data format.

### DIFF
--- a/docs/layer_util.js
+++ b/docs/layer_util.js
@@ -169,8 +169,8 @@ function addImageLayer(map, imageAsset, layer) {
     visParams: imgStyles,
     callback: (layerId, failure) => {
       if (layerId) {
-        layerArray[index].overlay = addLayerFromId(map, layerId, index,
-            layerArray[index].displayed);
+        layerArray[index].overlay =
+            addLayerFromId(map, layerId, index, layerArray[index].displayed);
       } else {
         // TODO: if there's an error, disable checkbox, add tests for this.
         layerArray[index].displayed = false;
--- a/docs/layer_util.js
+++ b/docs/layer_util.js
@@ -122,19 +122,20 @@ function getColorOfFeature(feature) {
  * should be able to call {@code map.overlayMapTypes.setAt(...)}.
  *
  * @param {google.maps.Map} map
- * @param {ee.Element} layer
- * @param {string} layerName
+ * @param {ee.Element} imageAsset
+ * @param {Object} layer
  * @param {number} index
  */
-function addImageLayer(map, layer, layerName, index) {
-  const imgStyles = firebaseLayers[layerName]['vis-params'];
-  if (firebaseLayers[layerName]['use-terrain-style']) {
-    layer = terrainStyle(layer);
+function addImageLayer(map, imageAsset, layer, index) {
+  const imgStyles = layer['vis-params'];
+  if (layer['use-terrain-style']) {
+    imageAsset = terrainStyle(imageAsset);
   }
   // Add a null-overlay entry to layerMap while waiting for the callback to
   // finish.
+  const layerName = layer['ee-name'];
   layerMap[layerName] = new LayerMapValue(null, index, true);
-  layer.getMap({
+  imageAsset.getMap({
     visParams: imgStyles,
     callback: (layerId, failure) => {
       if (layerId) {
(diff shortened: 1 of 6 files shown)


docs/layer_util.js
```diff
@@ -17,56 +18,88 @@ export {
...
 let deckGlOverlay;
 
-/** Values of layerMap. */
-class LayerMapValue {
+/**
+ * Values of layerArray. In addition to basic data from constructor, will
+ * have a data attribute for deck layers, and an overlay attribute for
+ * image/other layers.
+ */
+class DisplayedLayerData {
```
juliexxia: Class name is somewhat confused since the layer is not always displayed right?

docs/layer_util.js
```diff
@@ -17,56 +18,88 @@ export {
...
 
+/**
+ * Utility class to group deck parameters together, and cache computed color
+ * function.
+ */
+class DeckParams {
+  /**
+   * @constructor
+   *
+   * @param {!string} deckId
```
juliexxia: can you add description for this param

docs/layer_util.js
```diff
@@ -80,69 +113,65 @@ function setMapToDrawLayersOn(map) {
...
- * Function object to extract a color from a JSON Feature.
- *
- * @param {GeoJSON.Feature} feature
- * @return {Array} RGBA array
- */
-function getColorOfFeature(feature) {
-  return showColor(feature.properties['color']);
+function toggleLayerOff(index, map) {
+  const layerMapValue = layerArray[index];
+  layerMapValue.displayed = false;
```
juliexxia: nit: layerArrayValue ? think there's a couple of these

docs/layer_util.js
```diff
@@ -80,69 +113,65 @@ function setMapToDrawLayersOn(map) {
...
-function toggleLayerOn(layerName, map) {
-  const currentLayerMapValue = layerMap.get(layerName);
-  currentLayerMapValue.displayed = true;
-  if (currentLayerMapValue.data) {
-    addLayerFromFeatures(currentLayerMapValue, layerName);
+function toggleLayerOn(layer, map) {
+  const index = layer['index'];
+  const layerMapValue = layerArray[index];
+  layerMapValue.displayed = true;
+  if (layerMapValue.data) {
```
juliexxia: Is this still the right check for image layers? Am I understanding correctly that those have an overlay attr instead of a data attr? Do we not cache image layers?

docs/layer_util.js
```diff
@@ -195,20 +224,25 @@ function addLayerFromId(map, layerName, layerId, index, displayed) {
  * explicitly does not check the actual data. That's why the score layer needs
  * special handling, so deck.gl is forced to re-render it on parameter changes.
  *
- * @param {LayerMapValue} layerMapValue
- * @param {string} layerName
+ * @param {DisplayedLayerData} layerMapValue
+ * @param {number|string} index index of layer. A number unless this is the
+ * score layer
```
juliexxia: I think in the future we might also treat the SVI and SNAP layers like the score layer (index-less) if we go the route of not needing them to be explicitly specified as layer assets. Just a thought.

docs/layer_util.js
```diff
@@ -195,20 +224,25 @@ function addLayerFromId(map, layerName, layerId, index, displayed) {
...
  */
-function addLayerFromFeatures(layerMapValue, layerName) {
-  layerArray[layerMapValue.index] = new deck.GeoJsonLayer({
-    id: layerName,
+function addLayerFromFeatures(layerMapValue, index) {
+  const deckParams = layerMapValue.deckParams;
+  if (!deckParams.colorFunction) {
```
juliexxia: When will this trigger?

docs/layer_util.js
janakdr: Yeah, I'm trying to say that this is data about how a layer is actually drawn on the map. Would you prefer DrawnLayerData? RenderedLayerData?

docs/layer_util.js
janakdr: Very nice catch! I realized this last night after mailing for review, but it's not a new bug in this PR, so I was going to fix it in a follow-up. Added a TODO.

docs/layer_util.js
janakdr: renamed to displayedLayerData, can rename again if you want a different name for the class

docs/layer_util.js
janakdr: why would they be index-less? They would still have a relative position in the rankings, no? The score is special because it's always last.

docs/layer_util.js
janakdr: The same time getStyleFunction triggered createStyleFunction before: when this layer is first rendered, and colorFunction hasn't been computed yet.

juliexxia: Ah, gotcha, was mixing this up with colorFunctionProperties. So this is an attribute that is only read and accessed in this method?

docs/layer_util.js
juliexxia: Maybe LayerDisplayData? I like that slightly more.

docs/layer_util.js
janakdr: And when we create the score DeckParams object.

docs/layer_util.js
janakdr: Yeah, we can auto-populate that in the disaster creation, but I think the map viewer can be agnostic about that?

juliexxia:
No! LGMT, I'll approve now so you can submit after rename.

PR #210, opened 2019-11-15
add cancel button

juliexxia:
Add a button to cancel an add disaster process. Also disable and set the disaster picker to an option that just displays '...'.

### DIFF
--- a/docs/import/add_disaster.html
+++ b/docs/import/add_disaster.html
@@ -27,7 +27,7 @@
   </select>
   <span> </span>
   <button type="button" id="create-new-disaster">Create New Disaster</button>
-</form><p></p>
+</form><br>
 <div id="status"></div>
 <div id="new-disaster" hidden="true">
   <button type="button" id="cancel-new-disaster">Cancel</button>
--- a/docs/import/add_disaster.html
+++ b/docs/import/add_disaster.html
@@ -20,14 +20,15 @@
 </head>
 <body>
 
-<form action="" id="disaster-picker">
+<div id="disaster-picker">
   <label for="disaster">Disaster: </label>
-  <select id="disaster" style="width: 160px">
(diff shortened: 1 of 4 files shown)


docs/import/add_disaster.js
```diff
@@ -120,13 +118,30 @@ function writeNewDisaster(disasterId, states) {
   });
   if (!added) $('#disaster').append(createOptionFrom(disasterId));
   $('#disaster').val(disasterId);
+  disableDisasterPicker(false);
```
janakdr: nit: can you rename this to enableDisasterPicker? Search internally for "Improving Readability With Positive Booleans"

docs/import/add_disaster.html
```diff
@@ -22,12 +22,15 @@
 
 <form action="" id="disaster-picker">
   <label for="disaster">Disaster: </label>
-  <select id="disaster" style="width: 160px"></select>
+  <select id="disaster" style="width: 160px">
```
janakdr: Why not just hide the whole select element? I think you can put it inside a <span> that keeps its width to avoid any jumping around. Are the dots actually pulling their weight?

juliexxia: I've (not very tidily) changed it so it hides the disaster picker. Personally I like how the other way looks better, I think it flows more smoothly when the whole page doesn't change. But lmk if you have strong feelings otherwise!

docs/import/add_disaster.js

docs/import/add_disaster.html
janakdr: Why not implement my suggestion of wrapping the select in an element that keeps its width, so nothing else moves? Not on a computer right now, but sounds like that's not what you did?

juliexxia:
An approval and a passing travis do mine eyes deceive me

Issue #177, opened 2019-10-29
Support non-ee asset storage


ruthtalbot:
- Possible data like .kml files and map tiles

Issue #173, opened 2019-10-29
Improve image collection upload page UI


ruthtalbot:
- Update image collection upload page to be cleaner

janakdr:
Given that image upload is hacky, and tiles are just better and available in a lot of cases, I think we shouldn't invest more time in this. @ruthtalbot ?

ruthtalbot:
Agreed, but I'd still like to preserve the flow (if it works at all?), just in case. I can write it up in the documentation just in case Image Collections are ever needed.

janakdr:
It's currently completely broken. What priority do you think it should be to fix?

ruthtalbot:
Definitely below everything we've listed besides perhaps the link creation. 

Havent looked into it enough to know if it's applicable, but looking at https://gdal.org/programs/gdal2tiles.html to see if maybe we can just make them convert image collections to tiles/kmls in the worst case?

janakdr:
The problem with generating map tiles is they have to be hosted somewhere, and that won't be free. EE is slow, but free.

PR #215, opened 2019-11-16
Support collections of map tile URLS as layers

janakdr:
NOAA imagery is available as map tiles, which avoids the issue of having to upload images to EarthEngine. Because the user will want to view/hide all the tiles as a set, we implement our own custom MapType, similar to ImageMapType, but able to handle a collection of URLS. It fetches all of them, and for all the ones that returned images, stacks them on top of each other and returns the whole stack (inside a div).

There is a regression in this change, in that we no longer know when the tiles have finished rendering. EarthEngine's version of this overlay received a tileEvent that let it know when it was done rendering all tiles. I'm not sure how to implement that just yet, but that can be a follow-up, especially for after our dry run.

I've added a map tile asset in the new "layers" array in Firestore, and switched to using "layers" in this PR. It's not a huge problem if we switch to using "layers" before this PR goes in: there will be a console error if you try to enable the "NOAA Tiles" layer, but everything else still works.

This isn't actually work towards #173, but it makes that issue much less urgent.

Work towards #177


janakdr:
Actually, #219, the follow-up to this, changes enough of the internals that it's probably not worth reviewing them separately. Closing this one, unless you feel like #219 is so complex that you want to start by reviewing this.

PR #216, opened 2019-11-16
Stub out window.confirm better

janakdr:
Turns out we can just use Cypress stubs. Who knew!

### DIFF
--- a/docs/import/add_disaster.js
+++ b/docs/import/add_disaster.js
@@ -39,7 +39,7 @@ function enableWhenReady() {
 
   const deleteButton = $('#delete');
   deleteButton.prop('disabled', false);
-  deleteButton.on('click', () => deleteDisaster(window));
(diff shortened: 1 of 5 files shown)


docs/import/add_disaster.js
```diff
@@ -284,14 +284,12 @@ function createAssetPickers(states) {
 /**
  * Deletes a disaster from firestore. Confirms first. Returns when deletion is
  * complete (or instantly if deletion doesn't actually happen).
- * @param {Window} globalWindow DOM Window (for test injection)
  * @return {Promise<void>}
  */
-function deleteDisaster(globalWindow) {
+function deleteDisaster() {
```
juliexxia: there's a ref to this method further up in enableWhenReady do you mind getting rid of the param there as well. Thanks!

PR #221, opened 2019-11-17
Retry test specs on failure in Travis

janakdr:
Use the semi-supported cypress-run module to auto-retry failed specs. This is mostly for the mysterious (presumably Firebase-related) "Script errors" that affect unit_tests/{polygon_draw,add_disaster,center}_test.js. An example of a successful retry is https://travis-ci.com/givedirectly/Google-Partnership/builds/136887423. Unfortunately, it shows up as a failure on the Cypress dashboard (https://dashboard.cypress.io/#/projects/jr8ks8/runs/879/specs), but Travis shows it pass.

Also try to deflake integration_tests/polygon_draw_test.js a bit more. That was the only non-unit-test failure I noticed when testing this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,4 +41,5 @@ before_script:
   - yarn run ws --directory docs --port 8080 &
 script:
 # Use cypress-run package to auto-retry failed tests 3 times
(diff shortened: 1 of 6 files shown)


janakdr:
Think I actually figured out the root cause of the flakiness. Closing this in favor of #222. We can re-open if there are further unsolvable problems.

PR #222, opened 2019-11-17
Enforce proper script loading order in script_loader.js

janakdr:
Adding all the Firebase scripts to the page right away is dangerous, because I guess the browser can choose to load them in whatever order it wants. Instead, make sure the basic script is loaded first, and the other ones come after. Tested by running the problematic unit tests in a loop on Travis. Only got failures on the control: 
failuresOut of many runs: 
image and 
image

Also slightly deflake integration_tests/polygon_draw_test.js.

### DIFF
--- a/cypress/support/script_loader.js
+++ b/cypress/support/script_loader.js
@@ -3,48 +3,55 @@ import {cypressTestCookieName, earthEngineTestTokenCookieName, firebaseTestToken
 
 export {addFirebaseHooks, loadScriptsBeforeForUnitTests};
 
+/**
+ * Scripts that unit tests may want to load. Values have script and callback
+ * attributes, and optionally an extraScripts and extraCallback attributes, when
+ * they must load additional scripts that depend on the first one. More
+ * complicated graphs are not currently supported!
+ * @type {Map<string, Object>}
+ */
 const scriptMap = new Map([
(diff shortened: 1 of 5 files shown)


cypress/support/script_loader.js
```diff
@@ -87,7 +97,18 @@ function loadScriptsBeforeForUnitTests(...scriptKeys) {
...
+      const cypressPromise = waitForCallback(callback);
+      // If this script had dependent scripts, load them when this script has
+      // been loaded, and add the callback for those scripts in.
+      const scriptsAndCallbacks = extraScripts.get(callback);
+      if (scriptsAndCallbacks) {
+        cypressPromise.then(() => {
+          for (const script of scriptsAndCallbacks.scripts) {
+            addScriptToDocument(script);
+          }
+          return waitForCallback(scriptsAndCallbacks.callback);
```
juliexxia: why return here?

janakdr: As we've discussed, I like returning promises/chainables when available. If someone adds another then clause after this, they won't have to change this line to return. Without the return, it would be a very subtle gotcha, because the next then wouldn't wait for the first then to complete. Avoiding that bug seems worth returning a useful value when available.

PR #224, opened 2019-11-18
Change disaster name to include the year

janakdr:
This is an incompatible change. I renamed the EarthEngine folders to have the right form and modified Firestore data to point to the new assets.

Most of the work in resources.js is wasted, because we'll be deleting it in favor of Firestore-based disaster metadata retrieval, but whatever.

### DIFF
--- a/docs/resources.js
+++ b/docs/resources.js
@@ -1,5 +1,5 @@
-import {inProduction} from './in_test_util.js';
 import {gdEePathPrefix} from './ee_paths.js';
+import {inProduction} from './in_test_util.js';
 
 export {getDisaster, getResources};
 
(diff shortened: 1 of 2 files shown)


docs/resources.js
```diff
@@ -51,25 +50,30 @@ class DisasterMapValue {
...
+
 disasters.set(
-    'michael',
+    michaelName,
     new DisasterMapValue(
-        'michael', 2018, 'users/gd/michael/FEMA_Damage_Assessments',
-        'users/gd/michael/snap', 'users/gd/michael/tiger',
-        'users/gd/michael/income', 'users/gd/michael/svi',
-        'users/gd/michael/relevant_buildings'));
+        michaelName, michaelPathPrefix + 'FEMA_Damage_Assessments',
```
juliexxia: Maybe add a TODO to not store the path prefix for each of these since we can infer it form the disaster name?

janakdr: I think that we can't in all cases (because we might be reading directly from the state table, right?). But in any case, this will be mostly made moot by the previous TODO, that all this data will be read in from Firestore. If we want to store the data in Firestore without prefixes, I guess we can, but not sure it's worth it for the potential confusion.

Issue #185, opened 2019-10-31
Add 'add/edit asset' page


ruthtalbot:

juliexxia:
I think we can close this with the advent of manage_layers.js ?

PR #226, opened 2019-11-19
Table

juliexxia:
Included in this PR: 
* reading of all layers in firestore
* draggable layer order will propagate reorders to firestore
* display on load toggling with propagate to firestore

Future Work:
* Editable display name
* Editable color function
* Better support for non EE Feature collection typed assets (Image Collection viz params, KML urls, etc)
* Asset add/delete functionality


juliexxia:
Let me know what y'all think of the general layout and if all sounds good I'll add tests. 

image

juliexxia:
image

> Can we move the 'add additional assets..' to the bottom of the page?

Moved down and added some breaks - lmk if this is what you were thinking!

> I find having the 'Create New Disaster' right next to the dropdown a little unintuitive. It feels like you should select something from the dropdown then click 'Create New Disaster' not one or the other. Wondering if we can solve this by adding and 'OR' in between them or putting 'Create New Disaster' as an option in the dropdown list. Might not even be an issue for Alex though, so I can check with him tomorrow.

The create new disaster was originally an option in the drop down list but janak brought up the point that it created some unneeded complexity. Added that 'or' wdyt? It's a little wonky but not too bad I think.

> Is it difficult to display the first kml/map tile url somewhere? Either the same column as asset path or elsewhere. We'll definitely want those urls to be updatable post fact (path of the storm can change multiple times a day etc), so should probably surface it somewhere.
> -Is it possible to surface map tiles/kml under 'Types' as of now?

Surfacing an example - I think there's some cleanup to do between MAP TILES and KML but it'll generally look like this

docs/import/add_disaster.html
```diff
@@ -15,10 +15,13 @@
   <script src="https://www.gstatic.com/firebasejs/7.2.1/firebase-auth.js"></script>
   <!-- Load jQuery, a 3rd-party library. -->
   <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.4/jquery-ui.js"></script>
+
   <script src="https://apis.google.com/js/api.js"></script>
   <script type="module" src="./add_disaster_startup.js"></script>
 
   <!-- Load our styles. -->
+  <link rel="stylesheet" href="./style.css">
```
janakdr: I think later stylesheets override earlier ones. Might be better to put the page-specific css last?

docs/import/add_disaster.js
```diff
@@ -60,14 +63,152 @@ function enableWhenReady() {
...
+      .doc(currentDisaster)
+      .set(
+          {layerArray: disasterData.get(currentDisaster)['layerArray']},
+          {merge: true});
+}
+
+/**
+ * Update the table and disasterData with new indices after a layer has been
+ * reordered. Then write to firestore.
+ * @param {Object} ui
```
janakdr: nit: document this parameter?

docs/import/add_disaster_startup.js
```diff
@@ -26,4 +25,19 @@ authenticator.start();
 $('#create-new-disaster').on('click', () => toggleState(false));
 $('#cancel-new-disaster').on('click', () => toggleState(true));
 
+$('#tbody').sortable({
+  revert: true,
+  update: (event, ui) => updateAfterSort(ui),
+  // TODO: Table is still doing some jumping around on resort.
+  helper: function(e, tr) {
```
janakdr: nit: does this need to be a function as opposed to a lambda?

docs/import/add_disaster_startup.js
```diff
@@ -26,4 +25,19 @@ authenticator.start();
...
+    const originals = tr.children();
+    const helper = tr.clone();
+    helper.children().each(/* @this HTMLElement */ function(index) {
+      // Set helper cell sizes to match the original sizes
```
janakdr: nit: might be more helpful for this comment to be at the helper: ... line, and to explain what it's for:
"Keep each column in the dragged row the same width, even while it's being dragged" or similar?

docs/import/add_disaster.js
```diff
@@ -60,14 +63,152 @@ function enableWhenReady() {
...
+
+    // type
+    row.append(createTd(layerTypeStrings.get((layer['asset-type']))));
+
+    // display on load
+    const displayCheckbox =
+        $(document.createElement('input'))
+            .prop('type', 'checkbox')
+            .prop('checked', layer['display-on-load'])
+            .on('change', () => {
```
janakdr: It feels to me like this manual syncing of form<->data<->Firestore is going to get pretty repetitive. My suspicion is that something like AngularJS might be what the cool kids would do for a form that has to stay synced to a backend. Angular's blurb says in part: "Data-binding is an automatic way of updating the view whenever the model changes, as well as updating the model whenever the view changes. This is awesome because it eliminates DOM manipulation from the list of things you have to worry about."

I can imagine that we won't have time to learn/integrate Angular before the run-through, and so maybe manual coding of change handlers is the way to go, at least for now. What do you think?

If we do code change handlers ourselves, it sounds like jquery can attach a change handler to every element of a given class, so I think we should try to do that. We can then manually read the data from the row of the given changed element, which will be a little bit of unnecessary reading, and then write the whole array back to Firestore. Something like:

$('input.firestore_editable').change(function () {
   // Here, $(this) refers to the specific element of the .firestore_editable class which 'changed'
   updateLayerDataAtIndex(getIndexOfElement($(this)));
   return writeDataToFirestore();
}):

I'd recommend doing in this PR, so we know that it will work properly, because if that won't work, I would want to know alternatives to coding up 10 different change handlers before we proceed with this approach. But I'll leave up to you.

docs/import/add_disaster.js
```diff
@@ -60,14 +63,152 @@ function enableWhenReady() {
...
+  if (LayerType.hasOwnProperty(t)) {
+    layerTypeStrings.set(LayerType[t], t);
+  }
+}
+
+/**
+ * Write the current state of {@code disasterData} to firestore.
+ * @return {Promise<void>} Returns when finished writing.
+ */
+function writeDataToFirestore() {
```
janakdr: I think we need a queueing mechanism here similar to the one in StoredShapeData#update. Right now, every change kicks off an independent Firestore write. That means that the writes can race, so newer data could be overwritten by older data. The way to fix this (at least, the way it's done in StoredShapeData#update) is to keep track of the current state, Either SAVED, WRITING, or QUEUED_WRITE. Then when a write comes in, you either write it immediately, transitioning to WRITING state (if in SAVED state), do nothing (if in QUEUED_WRITE mode), or go from WRITING to QEUED_WRITE mode but do nothing else.

Then, when a write finishes, if that thread sees that we're in QUEUED_WRITE mode, it changes the mode back to WRITING and writes again. This ensures that we only have one write in flight at any given time, but will write new data as soon as it's safe to do so.

janakdr:
My preference re the "create new disaster" vs list thing was to have a newline (plus an optional <hr> or similar), but Julie wasn't a fan.

docs/import/add_disaster.js
janakdr: Hmm, I think my suggestion isn't great because the entries for each row will vary based on data type. So it might be easier to trigger just the relevant data mutation, since the input being modified will have a good idea of what data needs to change. Maybe at least we can extract the JavaScript handlers to standalone functions that are all next to each other, so it's easier to keep track of them.

docs/import/add_disaster.html
```diff
@@ -35,8 +38,26 @@
...
   <button type="button" id="create-new-disaster">Create New Disaster</button>
-</div>
+</div><br>
 <div id="status"></div>
+<div id="selected-disaster" hidden="true">
+  <table id="layer-manager" class="table table-bordered pagin-table">
+    <thead>
+      <tr>
+        <th></th>
+        <th style="width:200px">Asset</th>
```
ruthtalbot: Can we update 'Asset' to be 'Layer Name' or 'Layer Display Name'?

docs/import/add_disaster.js
```diff
@@ -60,14 +63,152 @@ function enableWhenReady() {
...
+    // TODO: make this editable.
+    row.append(createTd(layer['display-name']));
+
+    // asset path/url sample
+    // TODO: consolidate kml-urls and urls when we have a consistent naming
+    // convention.
+    if (layer['ee-name']) {
+      row.append(createTd(layer['ee-name']));
+    } else if (layer['urls']) {
+      row.append(createTd(layer['urls'][0]));
```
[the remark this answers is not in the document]
juliexxia: all consolidated to urls now

docs/import/add_disaster_startup.js
juliexxia: Done.

docs/import/add_disaster.js

docs/import/add_disaster.html
juliexxia: Done.

docs/import/add_disaster.html
juliexxia: Done.

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+ * @param {Object} layer
+ * @param {string} property
+ * @return {JQuery<HTMLElement> }
+ */
+function withCheckbox(td, layer, property) {
+  const checkbox = $(document.createElement('input'))
+                       .prop('type', 'checkbox')
+                       .prop('checked', layer[property])
+                       .on('change', (event) => {
+                         const index = $('#tbody > tr').length -
```
janakdr: This seems like a pretty indirect way to find the index (and it also seems expensive, since I'm guessing the $('tr').index below has to do at least a linear search).

Why not use the index-td of this row directly?

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+                             $('tr').index($(event.target).parents('tr'));
+                         getCurrentLayers()[index][property] =
+                             checkbox.is(':checked');
+                         return updateLayersInFirestore();
+                       });
+  return $(td).append(checkbox);
```
janakdr: why does td have to be wrapped here?

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+ * @param {JQuery<HTMLElement>} td
+ * @param {Object} layer
+ * @param {string} property
+ * @param {number} index
+ * @return {JQuery<HTMLElement>}
+ */
+function withColor(td, layer, property, index) {
+  const colorFunction = layer[property];
+  if (!colorFunction) {
+    td.html('N/A').addClass('na');
```
janakdr: I think we'll appreciate the consistency of keeping all data at the same HTML hierarchical level, so this N/A and class na would go on a span that was in the td, rather than at the td level. But we can change that in the next PR if you prefer (basically, as soon as you make this editable, I think it will make more sense to have the current status in a separate element that you can easily detach). My recommendation would be to have a single object inside the td. Either a color box, or a div containing multiple color boxes, or a div containing text. Then at the end of this function you append whatever element it is to the td. But my main concern is less the code duplication of td.append and more the issue of cleanly modifying this td without needing to check on what it's current state is.

Also, is it really true that it's N/A in this case? What if this is definitely a type that requires a color function, but the user hasn't entered it yet? I think we're going to want the user to be able to add a FeatureCollection without making them necessarily specify the color right away. Or are you envisioning that we always populate a default color?

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+  } else if (colorFunction['single-color']) {
+    td.append(createColorBox(colorFunction['single-color']));
+  } else if (colorFunction['base-color']) {
+    td.append(createColorBox(colorFunction['base-color']));
+  } else if (colorFunction['colors']) {
+    const colorObject = colorFunction['colors'];
+    const colorSet = new Set();
+    Object.keys(colorObject).forEach((propertyValue) => {
+      const color = colorObject[propertyValue];
+      if (!colorSet.has(color)) {
```
janakdr: nit: Save a line by doing if (colorSet.add(color)) {...}, since add returns whether or not the add happened.

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+        colorSet.add(color);
+        td.append(createColorBox(colorObject[propertyValue]));
+      }
+    });
+  } else {
+    console.log('layer ' + index + ': unrecognized color function');
```
janakdr: Can we have a TODO here for better error surfacing? Since we have setStatus, why not use it? And maybe make this console.error if not?

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+    // TODO: make this editable.
+    row.append(withText(createTd(), layer, 'display-name'));
+    // asset path/url sample
+    const assetOrUrl = createTd();
+    if (layer['ee-name']) {
+      withText(assetOrUrl, layer, 'ee-name');
+    } else if (layer['urls']) {
+      withList(assetOrUrl, layer, 'urls');
+    } else {
+      console.log('layer ' + i + ': unrecognized type');
```
janakdr: same here re console.log.

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
  * On change method for disaster picker.
  * @param {String} disaster
  * @return {Promise<void>} completes when we've finished filling all state
  * pickers and pulled from firebase.
  */
 function toggleDisaster(disaster) {
-  const states = disasters.get(disaster);
+  currentDisaster = disaster;
+
+  // display layer table
```
janakdr: nit: can we pull this out to a separate function so that toggleDisaster stays a bit more compact?

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+    row.append(withType(createTd(), layer, 'asset-type'));
+    // display on load
+    row.append(withCheckbox(createTd(), layer, 'display-on-load'));
+    // color
+    // TODO: make this editable.
+    row.append(withColor(createTd(), layer, 'color-function', i));
+    tableBody.append(row);
+  }
+
+  // display state asset pickers
```
janakdr: nit: can also pull this out to a separate function for symmetry.

docs/import/add_disaster.js
```diff
@@ -325,3 +563,29 @@ function createOptionFrom(innerTextAndValue) {
       .val(innerTextAndValue)
       .prop('id', innerTextAndValue);
 }
+
+/**
+ * Utility function for getting current data. May return undefined if
```
janakdr: when is this called before currentDisaster is set? If it shouldn't be, then just say that?

docs/import/add_disaster.js
```diff
@@ -325,3 +563,29 @@ function createOptionFrom(innerTextAndValue) {
...
+/**
+ * Utility function for getting current layers.
+ * @return {Array<Object>}
+ */
+function getCurrentLayers() {
+  return getCurrentData()['layers'];
+}
+
+/**
+ * Sets the current disaster so getCurrentData works for testing.
```
janakdr: I think this page should probably be disaster-setting aware. Ram's PR puts the current disaster name into localStorage. So you can use that in this PR and have Cypress set the value in localStorage.

juliexxia: Discussed offline - will handle in a follow up PR. Added a TODO.

juliexxia:
Now with tests!

docs/import/add_disaster.js
```diff
@@ -143,41 +157,43 @@ function updateAfterSort(ui) {
 
   // dumbly renumber all the layers. note: nth-child starts at 1, .index()
```
janakdr: nit: not so dumb anymore :) could stand a comment here explaining thinking

docs/import/add_disaster.js
janakdr: ugh, the worst. my bad

docs/firebase_layers.js
```diff
@@ -10,6 +10,7 @@ const LayerType = {
   IMAGE: 2,
   IMAGE_COLLECTION: 3,
   MAP_TILES: 4,
+  KML: 5,
```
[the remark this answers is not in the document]
janakdr: Might as well, so we're cleaner. Although may be moot if Ruth gets her PR in first :)

docs/import/add_disaster.js
```diff
@@ -60,14 +76,235 @@ function enableWhenReady() {
...
+      .doc(currentDisaster)
+      .set({layers: getCurrentLayers()}, {merge: true})
+      .then(() => {
+        pendingWriteCount--;
+        const oldState = state;
+        state = STATE.SAVED;
+        switch (oldState) {
+          case STATE.WRITING:
+            loadingElementFinished(writeWaiterId);
+            return Promise.resolve();
```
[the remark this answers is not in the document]
janakdr: It doesn't break the covenant of the doc. We are literally returning a promise, because the chain of functions that are actually executed returns a promise. This return is inside a lambda that is passed to then(). If the Firestore fetch never finished (for instance) this line of code would never be reached, even though this function would have long ago returned its Promise.

The .then() method of a Promise returns another Promise, let's call it thenPromise. What's inside thenPromise depends on what you return in your lambda. If you return a value, thenPromise will contain that value. If you return a Promise, thenPromise will contain whatever value was inside that Promise, when your returned Promise completes.

So, inside a then, returning a resolved Promise is exactly the same as returning the value that was inside that resolved Promise.

docs/import/add_disaster.js
janakdr: I'd rather surface it to the user, just like we surface illegal states. We can say: "Something went wrong, please contact developer. Include the message: '<>'". I think of this as "improving error messaging". We should try to avoid silent failures (from the user perspective).

docs/import/add_disaster.js
juliexxia: Ya juuuuust missed it. Updated.

docs/import/add_disaster.js
juliexxia: Gotcha - added something for now. Though there won't be a developer to contact in < two months...

docs/import/add_disaster.js
juliexxia: Also done here

docs/import/add_disaster.js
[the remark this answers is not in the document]
juliexxia: Ah,  because it's in a then. Sorry I misread.

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -242,8 +242,7 @@ describe('Unit tests for add_disaster page', () => {
     expect(discrete.children('.box').length).to.equal(2);
     expect(discrete.children().eq(1).css('background-color')).to.equal(red);
 
-    const log = cy.stub(console, 'log')
-                    .withArgs('layer 3: unrecognized color function');
+    const log = cy.stub(console, 'log').withArgs('unrecognized color function');
```
janakdr: do you still want this here? Shouldn't you be asserting on the status?

juliexxia: Yes, thanks

docs/import/add_disaster.js
```diff
@@ -64,7 +83,248 @@ function enableWhenReady() {
...
+        td.append(createColorBox(colorObject[propertyValue]));
+      }
+    });
+  } else {
+    setStatus(ILLEGAL_STATE_ERR + 'unrecognized color function: ' + layer);
+  }
+  return td;
+}
+
+/** Populates the layers table with layers of curernt disaster. */
```
janakdr: nit: s/curernt/current/

docs/import/add_disaster.js
```diff
@@ -64,7 +83,248 @@ function enableWhenReady() {
...
+    row.append(withColor(createTd(), layer, 'color-function', i));
+    tableBody.append(row);
+  }
+}
+
+/**
+ * Populates the state asset pickers with all known earth engine assets for
+ * those states.
+ * @return {Promise<void>} returns when all asset pickers have been populated
+ * after potentially retreiving states assets from ee.
```
janakdr: nit: s/retreiving/retrieving. And s/states/states'/

juliexxia:
Work towards #185, #176

PR #227, opened 2019-11-19
Persisting the selected disaster across page loads

RamV13:


docs/resources.js
```diff
@@ -12,14 +12,31 @@ function getResources() {
...
+  // default.
   const params = new URLSearchParams(window.location.search);
-  const disaster_param = params.has('disaster') ? params.get('disaster') : disaster;
-  return inProduction() ? disaster_param : disaster;
+  if (params.has('disaster')) {
+    final_disaster = params.get('disaster');
+  } else if (localStorage.getItem('disaster')) {
+    final_disaster = localStorage.getItem('disaster');
+  } else if (!inProduction()) {
+    // Start off with Harvey for test cases since our tests assert against
```
janakdr: I think we can get rid of this by having the tests just use the Harvey url? Or you could support a cookie, although I see that localStorage is nicer.

RamV13: I just set localStorage in Cypress instead.

cypress/integration/integration_tests/nav_test.js
```diff
@@ -6,6 +6,11 @@ describe('Integration test for the navbar', () => {
 
     // The page should now be switched to the Michael disaster.
```
janakdr: Do we want the URL-changing support at all, then?

RamV13: Removed

PR #229, opened 2019-11-20
Add support for Kml layers 

ruthtalbot:
Prototype of adding support for KML layers. Currently using an isArray check, but happy to add a .type to the layerDisplayData if wanted.


docs/layer_util.js
```diff
@@ -419,15 +423,40 @@ function addLayer(layer, map) {
           convertEeObjectToPromise(
               ee.FeatureCollection(layerName).toList(maxNumFeaturesExpected)),
           DeckParams.fromLayer(layer), layer['index']);
+      break;
+    case LayerType.KML:
+      addKmlLayers(layer, map);
```
janakdr: Nit: can you return a promise from this branch, even if it's resolved?

docs/layer_util.js
```diff
@@ -419,15 +423,40 @@ function addLayer(layer, map) {
...
+      break;
     case LayerType.MAP_TILES:
       return addTileLayer(map, layer);
+      break;
```
janakdr: Don't need to break after a return statement

docs/layer_util.js
```diff
@@ -150,7 +150,7 @@ function toggleLayerOn(layer, map) {
     addLayerFromFeatures(layerDisplayData, index);
     return null;
   }
-  if (layerDisplayData.overlay) {
+  if (layerDisplayData.overlay && !(Array.isArray(layerDisplayData.overlay))) {
```
janakdr: Can we at least make a utility function, either a method on LayerDisplayData or standalone, that says if it's kml, like, isKmlLayer? It would do the array check internally.

docs/layer_util.js
ruthtalbot: My bad, merged that incorrectly. Thanks!

PR #231, opened 2019-11-21
Remove calls to getResources()

janakdr:
Disasters are going to be retrieved live from Firestore, not hard-coded in resources.js, so we can't really depend on this data.

Also make our map tolerant to a missing damage layer. After this, if import_data produces a "score" asset with no damage, our map should work as expected.

### DIFF
--- a/docs/polygon_draw.js
+++ b/docs/polygon_draw.js
@@ -4,16 +4,18 @@ import {getFirestoreRoot} from './firestore_document.js';
 import {addLoadingElement, loadingElementFinished} from './loading.js';
 import {geoPointToLatLng, latLngToGeoPoint} from './map_util.js';
 import {createPopup, isMarker, setUpPopup} from './popup.js';
-import {getResources} from './resources.js';
+import {getScoreAsset} from './resources.js';
 import {userRegionData} from './user_region_data.js';
 
 // StoredShapeData is only for testing.
 export {
   displayCalculatedData,
-  processUserRegions,
+  initializeAndProcessUserRegions,
   setUpPolygonDrawing,
   StoredShapeData,
 };
+
+let damageAsset = null;
 
 /**
  * Class holding data for a user-drawn feature (marker or polygon), including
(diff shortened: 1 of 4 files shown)


docs/polygon_draw.js
```diff
@@ -349,15 +351,20 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
  *
  * @param {google.maps.Map} map Map to display regions on
- * @param {Promise<any>} firebasePromise Promise that will complete when
- *     Firebase authentication is finished
+ * @param {Promise<any>} firebasePromise Promise with Firebase damage data (also
+ *     implies that authentication is complete)
  * @return {Promise} promise that is resolved when all initialization is done
  *     (only used by tests).
  */
 function processUserRegions(map, firebasePromise) {
```
juliexxia: maybe update function name and doc? + maybe a '.' a the end of the firebasePromise param?

docs/resources.js
```diff
@@ -18,8 +18,13 @@ function getDisaster() {
...
   }
-  localStorage.setItem('disaster', default_disaster);
-  return default_disaster;
+  localStorage.setItem('disaster', defaultDisaster);
+  return defaultDisaster;
+}
+
+/** @return {string} EE asset path for score asset for the current disaster */
+function getScoreAsset() {
+  return gdEePathPrefix + getDisaster() + '/data-ms-as-nod';
```
juliexxia: this file name is now potentially bogus right? as in could be without any damage info at all. Not worth rerunning scripts but maybe worth leaving a note until it's changed.

docs/polygon_draw.js
janakdr: I don't think we have a convention of ending parameter docs with periods (see the previous line). Updated doc and renamed.

docs/resources.js
janakdr: import_data.js still will only generate this asset with damage, so I don't think it's necessary yet?

Also, I didn't parse "ms-as-nod" as "Microsoft-as-no-damage" until just now (thanks!), so may not be essential.

docs/polygon_draw.js
juliexxia: It's just so similar to the comment below, which does end in a period, but I won't die on this hill.

docs/resources.js
juliexxia: Agreed, actually rerunning this scripts not worth it and maybe fine if you're going to change it tomorrow but it might be nice to leave a note here indicating that what's returned by this method and what the asset path describes could be two different things.

docs/polygon_draw.js
janakdr: Removed period from @return statement :)

PR #232, opened 2019-11-21
Use Firestore disaster list in navbar

janakdr:
Also take the opportunity to refactor script.js to use our now standard model of tracking pending tasks.

With this, resources.js's disasters are only used by import_data.js, and will be cleaned up in a follow-up.


docs/navbar.js
```diff
@@ -15,12 +15,10 @@ function loadNavbar(callback) {
  * @param {string} title the title of the navbar
  */
 function loadNavbarWithTitle(title) {
-  $(() => {
-    $('#navbar').load('/navbar.html', () => {
+  $(() => loadNavbar(() => {
```
RamV13: Thanks :)

docs/disaster_picker.js
```diff
@@ -1,21 +1,31 @@
...
-    disasterDropdown.appendChild(disasterItem);
-  });
+/**
+ * Initializes the disaster picker.
+ * @param {Promise} firebaseAuthPromise Promise that completes when logged in to
+ *     Firebase
+ */
+function initializeDisasterPicker(firebaseAuthPromise) {
+  firebaseAuthPromise.then(() => {
+    const disasterDropdown = document.getElementById('disaster-dropdown');
```
juliexxia: jquery for consistency? Also the method createOptionFrom exists if you're interested.

docs/disaster_picker.js
```diff
@@ -1,21 +1,31 @@
...
+    getDisasters().then((querySnapshot) => {
+      querySnapshot.forEach((disasterDoc) => {
+        const disasterItem = document.createElement('option');
+        const disaster = disasterDoc.id;
+        disasterItem.innerHTML = disaster;
+        if (disaster === getDisaster()) {
```
juliexxia: maybe pull this out so it only gets called once?

cypress/plugins/index.js
```diff
@@ -119,15 +119,20 @@ module.exports = (on, config) => {
...
-            documentReference.set({dummy: true}, {merge: true}),
-          ]))
+      const writePromises = [];
+      for (const disaster of ['2017-harvey', '2018-michael']) {
+        const documentPath = 'disaster-metadata/' + disaster;
+        const prodDisasterDoc = prodApp.firestore().doc(documentPath);
+        const testDisasterDocReference = testApp.firestore().doc(
+            'test/' + currentTestRoot + '/' + documentPath);
+        writePromises.push(
+            Promise.all([prodDisasterDoc.get(), signinPromise, deletePromise])
```
juliexxia: Is it weird to wait on these twice in a sequential loop?

docs/disaster_picker.js
janakdr: Done on jQuery. Depending on add_disaster.js from this seems weird, and I don't want to refactor it out of add_disaster.js because it would mess with your PR (also the selects there have a bit more structure than the ones here need).

cypress/plugins/index.js
janakdr: They're not actually being waited on in the loop, they're just being put into a Promise.all. We then wait on them in parallel in the Promise.all below. So the graph of Promise waiting looks like:
image

Note that we construct that entire graph before we do any waiting, so there's no sequential waiting.

docs/disaster_picker.js
```diff
@@ -1,21 +1,32 @@
...
+ */
+function initializeDisasterPicker(firebaseAuthPromise) {
+  firebaseAuthPromise.then(() => {
+    getDisasters().then((querySnapshot) => {
+      querySnapshot.forEach((disasterDoc) => {
+        const currentDisaster = getDisaster();
+        const disasterDropdown = $('#disaster-dropdown');
+        const disasterItem = document.createElement('option');
+        const disaster = disasterDoc.id;
+        disasterItem.innerHTML = disaster;
```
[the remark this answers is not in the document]
janakdr: Done.

docs/disaster_picker.js
```diff
@@ -11,9 +11,9 @@ export {initializeDisasterPicker};
 function initializeDisasterPicker(firebaseAuthPromise) {
   firebaseAuthPromise.then(() => {
     getDisasters().then((querySnapshot) => {
+      const disasterDropdown = $('#disaster-dropdown');
       querySnapshot.forEach((disasterDoc) => {
         const currentDisaster = getDisaster();
```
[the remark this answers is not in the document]
janakdr: Done, thanks. jQuery all the things!

janakdr:
Followed suggestions, but can't actually respond because of rebases.

janakdr:
Had funny bug in script.js: was accidentally running all the "run" stuff twice. Fixed, will submit when tests pass.

janakdr:
darn it all, merged into wrong upstream branch :( incoming...

PR #234, opened 2019-11-21
Use local storage for current disaster for add_disaster.js and make name/lists editable

juliexxia:
Use cookie for current disaster for add_disaster.js


docs/import/add_disaster.js
```diff
@@ -603,7 +595,7 @@ function getCurrentLayers() {
  * @param {string} disasterId
  */
 function setCurrentDisaster(disasterId) {
```
janakdr: remove from exports?

juliexxia:
Didn't realize what branch I was on and pushed commits for making fields editable to this branch. Janak, let me know if it's too confusing to review all together. In the meantime, here's what the new changes look like:

image

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -261,55 +267,52 @@ describe('Unit tests for add_disaster page', () => {
 
   it('tests list cell', () => {
     const chocolate = 'chocolate';
+    const chai = 'chai';
+    const nutmeg = 'nutmeg';
+
+    setDisasterAndLayers([{flavors: [chocolate, chai]}]);
```
janakdr: Nit: extract object with flavors property to variable and use below?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -348,6 +351,43 @@ describe('Unit tests for add_disaster page', () => {
   });
 });
 
+/**
+ * Set's local cookie to point to disaster with the given layers.
```
janakdr: Nit: s/Set's/Sets/. Also not actually a cookie. "Local storage"?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -348,6 +351,43 @@ describe('Unit tests for add_disaster page', () => {
...
+  const currentDisaster = '2005-fall';
+  disasterData.set(currentDisaster, {layers: layers});
+  window.localStorage.setItem('disaster', currentDisaster);
+}
+
+/**
+ * Function that tests the save method works.
+ * @param {Function} fxn save function
+ * @param {string} property
+ * @param {Object} input DOM object save from which is pulling value
```
janakdr: Nit: can't parse this line

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -348,6 +351,43 @@ describe('Unit tests for add_disaster page', () => {
...
+ * @param {*} afterVal
```
janakdr: Nit: are we using * for param types?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -348,6 +351,43 @@ describe('Unit tests for add_disaster page', () => {
...
+      .then(() => {
+        expect(getCurrentLayers()[0][property]).to.eql(afterVal);
+        expect(loadingStartedStub).to.be.calledOnce;
+        expect(loadingFinishedStub).to.be.calledOnce;
+        return getFirestoreRoot()
+            .collection('disaster-metadata')
+            .doc(getDisaster())
+            .get();
+      })
+      .then((doc) => {
```
janakdr: Nit: can leave off braces

docs/import/add_disaster.js
```diff
@@ -172,6 +169,45 @@ function createTd() {
...
+ */
+function onUpdate(event, property, fxn) {
+  const input = $(event.target);
+  const index = input.parents('tr').children('.index-td').text();
+  getCurrentLayers()[index][property] = fxn(input);
+  return updateLayersInFirestore();
+}
+
+/**
+ * Auto save on focus out from input cell.
```
janakdr: Nit: "Auto-saves". Here and below

docs/import/add_disaster.js
```diff
@@ -191,7 +237,13 @@ function withText(td, layer, property) {
  * @return {JQuery<HTMLTableDataCellElement>}
  */
 function withList(td, layer, property) {
-  return td.text(layer[property][0]);
+  // use a regex to replace all instances of ',' with a new line
+  const textarea = $(document.createElement('textarea'))
+                       .val(layer[property].toString().replace(/,/g, '\n'))
```
janakdr: I don't think modifying values on display is the way to go. If we want to split based on commas (we can, but I'm not clear we need to), I think we should do it on value extraction.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: I don't think we have up until now, but this feels like a * situation - right now it's being used for a boolean, string and list.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Done.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Done.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Better?

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Done.

docs/import/add_disaster.js

docs/import/add_disaster.js
juliexxia: Not totally sure what you mean by "modifying values on display" - is this better?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -1,13 +1,17 @@
...
+import {addDisaster, createAssetPickers, createOptionFrom, createTd, deleteDisaster, disasterData, emptyCallback, getAssetsFromEe, getCurrentLayers, onCheck, onInputBlur, onListBlur, stateAssets, updateAfterSort, withCheckbox, withColor, withInput, withList, withType, writeNewDisaster} from '../../../docs/import/add_disaster.js';
 import * as loading from '../../../docs/loading.js';
+import {getDisaster} from '../../../docs/resources';
 import {addFirebaseHooks, loadScriptsBeforeForUnitTests} from '../../support/script_loader.js';
 
 const KNOWN_STATE = 'WF';
 const UNKNOWN_STATE = 'DN';
 const KNOWN_STATE_ASSET = gdEeStatePrefix + KNOWN_STATE + '/snap';
 
+let loadingStartedStub;
```
janakdr: nit: can you put this inside the describe?

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -261,55 +267,53 @@ describe('Unit tests for add_disaster page', () => {
 
   it('tests list cell', () => {
     const chocolate = 'chocolate';
```
janakdr: nit about these tests' style. It doesn't harm anything, but I also don't see the advantage of using variables for strings that are the same as the variables. You're just saving the single quotes? Especially below where you're joining the strings it might actually look neater to have them as literals.

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -261,55 +267,53 @@ describe('Unit tests for add_disaster page', () => {
...
-          expect(loadingFinishedStub).to.be.calledOnce;
-          return getFirestoreRoot()
-              .collection('disaster-metadata')
-              .doc(currentDisaster)
-              .get();
-        })
-        .then((doc) => {
-          expect(doc.data()['layers'][0][property]).to.be.true;
-        });
+    testSave(onCheck, property, checkbox, true);
```
janakdr: Can we add a test for uncheck?

docs/import/add_disaster.js
```diff
@@ -70,8 +68,7 @@ function enableWhenReady() {
     });
 
     disasterPicker.on('change', () => toggleDisaster(disasterPicker.val()));
-    const mostRecent = querySnapshot.docs[querySnapshot.size - 1].id;
-    disasterPicker.val(mostRecent).trigger('change');
+    disasterPicker.val(getDisaster()).trigger('change');
```
janakdr: Something to consider here is that we are only selecting the current disaster after we've loaded all of the Firestore data. So we could modify getDisaster() to take a list of known disasters if available. That way if no current disaster is selected in localStorage, it could return the latest disaster, rather than the default of 2017-harvey. Not necessary for this PR, but something to maybe keep in mind.

docs/import/add_disaster.js
```diff
@@ -172,6 +169,45 @@ function createTd() {
...
+ * A common update method that writes to local data and firestore based on
+ * a customizable version of the value of the input.
+ * @param {Object} event
+ * @param {string} property
+ * @param {Function} fxn how to read/transform the raw value from the DOM.
+ * @return {?Promise<void>} See updateLayersInFirestore doc
+ */
+function onUpdate(event, property, fxn) {
+  const input = $(event.target);
+  const index = input.parents('tr').children('.index-td').text();
```
janakdr: is this done anywhere else? Should we extract to a utility function?

docs/import/add_disaster.js
```diff
@@ -191,7 +237,17 @@ function withText(td, layer, property) {
  * @return {JQuery<HTMLTableDataCellElement>}
  */
 function withList(td, layer, property) {
-  return td.text(layer[property][0]);
+  const list = layer[property];
+  let withNewLines = list[0];
```
janakdr: why not list.join('\n')?

docs/import/add_disaster.js
```diff
@@ -191,7 +237,17 @@ function withText(td, layer, property) {
...
+  for (let i = 1; i < list.length; i++) {
+    withNewLines = withNewLines.concat('\n', list[i]);
+  }
+  // use a regex to replace all instances of ',' with a new line
```
janakdr: nit: comment stale?

docs/import/add_disaster.js
```diff
@@ -191,7 +237,17 @@ function withText(td, layer, property) {
...
+  const textarea =
+      $(document.createElement('textarea')).val(withNewLines).prop('rows', 3);
+  td.append(textarea);
+  td.on('blur', 'textarea', (event) => onListBlur(event, property));
```
janakdr: why not trigger on the textarea directly, rather than the td? Here and above with the input.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Isn't the advantage that you can't accidentally change one usage of the var without changing others? I thought generally good practice to variablize reused strings.

docs/import/add_disaster.js
juliexxia: Yep, better thanks

docs/import/add_disaster.js
juliexxia: Yep gone.

docs/import/add_disaster.js
juliexxia: Sure

docs/import/add_disaster.js
juliexxia: Extracted - not entirely sure it's worth it but maybe we'll use again.

docs/import/add_disaster.js
juliexxia: noted

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Sure. Added.

cypress/integration/unit_tests/add_disaster_test.js
juliexxia: Discussed offline.

cypress/integration/unit_tests/add_disaster_test.js
```diff
@@ -298,17 +298,29 @@ describe('Unit tests for add_disaster page', () => {
...
-    expect(checkbox.prop('checked')).to.be.false;
+    expect(checkbox.prop('checked')).to.eq(!checks);
 
-    checkbox.prop('checked', true);
-    testSave(onCheck, property, checkbox, true);
+    checkbox.prop('checked', checks);
+    testSave(onCheck, property, checkbox, checks);
+  }
+
+  it('tests checkbox cell check', () => {
```
janakdr: nit: can one-line, here and below :)

cypress/integration/unit_tests/add_disaster_test.js
janakdr: I'm less convinced of it in tests. Not a big deal either way

PR #236, opened 2019-11-22
Add table for selecting score data assets

ruthtalbot:


juliexxia:
I would sync to head! Might be helpful to see what it looks like with the table and you might pick up some css for your table. 

So the way it works now - all assets for all relevant states will be displayed here? That might get kind of hairy - it may make sense to do like a separate column per state so it's easy to tell if you've selected something in each category for each state - does that make sense?

docs/import/add_disaster.js
```diff
@@ -370,112 +376,76 @@ function populateStateAssetPickers() {
...
+  headerRow.empty();
 
-  if (assets.length) {
-    $('#score-asset-selection').show();
-  } else {
-    $('#score-asset-selection').hide();
+  // Initialize headers.
+  headerRow.append(createTd().html('Score Assets'));
+  for (const state of states) {
+    headerRow.append(createTd().html(state + ' Assets'));
```
juliexxia: Does this mean that the assets for different states need to share the same column names? I'm assuming this is fine because I can't imagine you're getting data about one state from one source and from another state from another source. Just making sure.

docs/import/add_disaster.js
```diff
@@ -370,112 +376,76 @@ function populateStateAssetPickers() {
...
+    for (let i = 0; i < scoreAssetTypes[scoreAssetType].length; i++) {
       columnCell.append($(document.createElement('input')).attr({
-        id: name + '-column-2',
+        id: scoreAssetType + '-column-' + i,
         type: 'text',
       }));
     }
+    row.append(columnCell);
+    tableBody.append(row);
+    row.change(() => handleScoreAssetSelection(scoreAssetType));
```
juliexxia: Intellij tells me that row.on('change', () => handeScore...) is the nondeprecated way to do this. Also, when exactly does this trigger? Is it if any elements inside the row change?

docs/import/add_disaster.js
```diff
@@ -370,112 +376,76 @@ function populateStateAssetPickers() {
...
+    for (const state of states) {
+      if (stateAssets.get(state)) {
+        row.append(createTd().append(createAssetDropdown(
+            stateAssets.get(state), scoreAssetType, state)));
+      }
+    }
+    // TODO: Make column names selects instead.
+    const columnCell = createTd();
+    for (let i = 0; i < scoreAssetTypes[scoreAssetType].length; i++) {
       columnCell.append($(document.createElement('input')).attr({
```
juliexxia: can we create labels for these inputs? Or like text that comes up if you hover? Helpful for the poverty asset I would think.

docs/import/add_disaster.js
```diff
@@ -370,112 +376,76 @@ function populateStateAssetPickers() {
...
-      .append($(document.createElement('input')).attr({
-        id: name + '-' + value,
-        type: 'radio',
-        name: name + '-select',
-        value: value,
-        checked: opt_checked
-      }));
+function createAssetDropdown(assets, row, state) {
+  // Create the asset selector and add a 'None' option.
+  const select = $(document.createElement('select')).attr({
```
juliexxia: can be slightly more concise with .prop('id', row + '-' + state) instead of the whole attr shebang

docs/import/add_disaster.js
```diff
@@ -41,6 +41,12 @@ const disasterData = new Map();
 // Map of state to list of known assets
 const stateAssets = new Map();
 
+const scoreAssetTypes = {
+  'Poverty': ['Total Households', 'SNAP Households'],
```
juliexxia: Are these values used anywhere?

docs/import/add_disaster.js
```diff
@@ -349,7 +355,8 @@ function populateStateAssetPickers() {
   // reloading the page.
   let assetPickersDone = Promise.resolve();
   if (statesToFetch.length === 0) {
-    createAssetPickers(states);
+    createAssetPickers(states, 'asset-pickers');
```
janakdr: nit: wrap these functions into one so we don't have to call them in turn twice?

docs/import/add_disaster.js
ruthtalbot: They're used to know how many columns to add, and I was thinking we could add them as labels to the columns in the future? But you asked about that above so I'll do it now :)

docs/import/add_disaster.js
ruthtalbot: Done!

docs/import/add_disaster.js
ruthtalbot: Done!

docs/import/add_disaster.js
ruthtalbot: Updated!

It triggers when anything int the row changes, which I figured we'd want -- do you want to make it more specific?

docs/import/add_disaster.js
ruthtalbot: Yeah, I talked about this with Janak at some point -- figured if someone is getting assets for poverty, etc they'd be from the same source and have the same column names.

docs/import/add_disaster.js
ruthtalbot: Sg, if we're probably going to be pulling this logic out into an import page should we keep them separate?

janakdr:
I'm interested in automating the creation of these pickers and column names based on the Firestore data. I'm also thinking that the column names should maybe be "below the fold," not in the same row as the assets, because we expect them never/rarely to have to modify them. What do people think of those two ideas? I think that if we go with that, then the only part of this PR we need to pursue is the state asset picker. Everything else can be done programmatically.

juliexxia:
Janak, are you anticipating an earth engine call to generate the column names every time the user changes the value of the asset picker? That might introduce a little bit of drag but I'm probably fine with that. I think makes sense if we're already generating columns for the color function stuff. 

Re: below or same line as the asset picker, no real opinion. I think same line keeps things tight but I'm fine with both.

docs/import/add_disaster.js
```diff
@@ -408,14 +407,16 @@ function initializeScoreSelectors(states) {
     // TODO: Make column names selects instead.
     const columnCell = createTd();
     for (let i = 0; i < scoreAssetTypes[scoreAssetType].length; i++) {
+      columnCell.append($(document.createElement('label'))
+                            .text(scoreAssetTypes[scoreAssetType][i]));
```
juliexxia: Might be nice to standardize the sizes of these label + input boxes so the boxes line up down the column. But this comment is pending our other discussion of whether or not we want to handle that in this PR.

janakdr:
I'm imagining that we pre-compute column names on load. Not sure where we
should keep the user from advancing, though. Maybe enable options in the
select as info becomes available?

docs/import/add_disaster.js
```diff
@@ -359,14 +366,88 @@ function populateStateAssetPickers() {
...
+ * @param {Array<string>} states array of state (abbreviations)
+ */
+function initializeScoreSelectors(states) {
+  const headerRow = $('#score-asset-header-row');
+  const tableBody = $('#asset-selection-table-body');
+  tableBody.empty();
+  headerRow.empty();
+
+  // Initialize headers.
+  headerRow.append(createTd().html('Score Assets'));
```
juliexxia: Where will picking the damage asset happen? and maybe we should rephrase Score Assets since income and svi aren't actually included in the score. But happy to leave the latter as a thought and see if it confuses Alex or not.

juliexxia:
So on page load we compute column names for all assets in all states? Which means we also ask for all assets for all states on load, which we're not doing now.

janakdr:
Yes, that seems reasonable to me. I was suggesting we do that for the "add
layer" functionality as well, so that it would hide any latency.

juliexxia:
Hm, ok, I think I'm fine with that.

janakdr:
I'm ok either way! Whatever's faster for this PR. It's going to change a
lot in the follow-up anyway.

docs/import/add_disaster.js
ruthtalbot: Yeah, I wasn't sure what to call them since they're in the table but not used in the score. If you have any idea lmk otherwise I'll brainstorm and change if I can come up with something better. 

For the damage asset, was picturing that based on conversations with you both as just a separate selector below that only shows assets in the disaster folder. Thoughts?

juliexxia:
Discussed offline with Ruth that picking the damage asset for the score would go in a separate table since not state based.

PR #237, opened 2019-11-22
Explicitly store color schema choice in firestore

juliexxia:
This will allow up to store more than one schema if the user is switching back and forth between color schemas. 

Along the way, rename STATE -> State to keep with other enums we've created. Also, section out a new color function file and a util file for add disaster because it's getting big and will probably continue to grow.


juliexxia:
Reminder to self - after submission, purge all instances of 'continuous' field from firestore.

PR #238, opened 2019-11-23
Make color functions editable via add_disaster.html

juliexxia:
Bring up a color function editor when you click on colors cell for a given row. 

This reads the color-function map from firestore and writes back to it every time the user makes a change. Also included is a one-off script (add_layer.js) to write the appropriate information to firestore but that's not actually integrated yet and will be in a follow up PR.

The code expects each colorable layer (i.e. feature collections) to have a map of column information that looks like column name -> max, min, [all, possible, values]. Running add_layer to generate this for a single feature collection took ~1-2 seconds including the firestore console refreshing itself. We may want to consider not including all possible values by default because for clearly non-discrete data, it's a lot of things to be storing in firestore that will never be used. But that does require some user pre-knowledge of the data, so I'm not sure.

Tests pending on general approval of the popup.


docs/import/color_function_util.js
```diff
@@ -1,8 +1,162 @@
...
+  updateLayersInFirestore();
+}
+
+/**
+ * Updates an individual color choice for a single value in the discrete schema.
+ * @param {JQuery<HTMLElement>} picker
+ */
+function setDiscreteColor(picker) {
+  const colorFunction = getColorFunction();
+  const propertyValue = picker.data('value');
```
janakdr: nit: pull this 'value' and 'value' from the picker creation out to a common constant?

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+/**
+ * Switches the schema of {@link globalTd} to the given type.
+ * @param {enum} type
+ */
+function switchSchema(type) {
+  const colorFunction = getColorFunction();
+  $('.color-type-div').hide();
+  switch (type) {
+    case ColorStyle.SINGLE:
+      $('#single-color-radio').prop('checked', true);
```
janakdr: nit: Do you actually need to do this? Isn't this triggered by the radio being checked?

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+      $('#continuous-radio').prop('checked', true);
+      $('#continuous-color-picker').val(colorFunction['color']);
+      populatePropertyPicker($('#continuous-property-picker'));
+      $('#continuous').show();
+      break;
+    case ColorStyle.DISCRETE:
+      $('#discrete-radio').prop('checked', true);
+      const propertyPicker = $('#discrete-property-picker');
+      populatePropertyPicker(propertyPicker);
+      if (propertyPicker.val()) populateDiscreteColorPickers();
```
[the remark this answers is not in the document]
juliexxia: Agreed, but sort of nice because it means we don't have to make a bunch of color pickers for discrete if this is a single-color data set. I can push it down into populateDiscreteColorPickers if that would help.

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+function populateDiscreteColorPickers() {
+  const pickerList = $('#discrete-color-pickers').empty();
+  const colorFunction = getColorFunction();
+  const values =
+      colorFunction['columns'][$('#discrete-property-picker').val()]['values'];
+  const asColorPickers = [];
+  const colors = getColorFunction()['colors'];
+  for (const value of values) {
+    const li = $(document.createElement('li'));
+    li.append($(document.createElement('label')).text(value + ': '));
```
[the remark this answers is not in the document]
juliexxia: No because these color pickers don't have ids so these labels aren't actually attached to them. I don't really want to give them ids because it sounds like overkill to try to construct unique ones just for attaching labels.

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+    const colorPicker =
+        createColorPicker()
+            .on('change', (event) => setDiscreteColor($(event.target)))
+            .data('value', value)
+            .val(colors ? colors[value] : null);
```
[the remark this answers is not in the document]
juliexxia: the colors map of value -> color (not to be confused with color, sorry probably not a great choice on my part) doesn't exist for an asset until you actually change one of the color pickers in the discrete div (see setDiscreteColor). Sort of nice because hopefully it will never exist for some assets that don't care about it.

docs/import/color_function_util.js
```diff
@@ -51,3 +319,12 @@ function createColorBox(color) {
       .addClass('box')
       .css('background-color', color);
 }
+
+/**
+ * Gets color function related to {@link globalTd}.
+ * @return {*}
```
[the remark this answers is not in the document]
juliexxia: Maybe better?

docs/import/add_disaster.html
```diff
@@ -58,6 +58,13 @@
     </div>
   </table>
   <br>
+  <div id="color-fxn-editor" class="editor" hidden="true">
```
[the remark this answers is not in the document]
juliexxia: It's now highlighted to the same green as the navbar which isn't suuuper highlighted but it's something. LMK what you think.

docs/import/add_disaster.js
```diff
@@ -50,6 +38,8 @@ const stateAssets = new Map();
  * @return {Promise<firebase.firestore.QuerySnapshot>}
  */
 function enableWhenReady() {
+  populateColorFunctions();
```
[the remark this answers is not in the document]
juliexxia: It does create some change handlers that do need firestore access/ read disasterData which we populate from firestore. So i think *really* it should go inside the fetch of getDisasters. On the other hand, I guess you can't access any of this until the layers table exists which itself waits on firestore. So sure.

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+    return;
+  }
+  globalTd = td;
+  $('#' + colorStyleTypeStrings.get(type) + '-radio')
+      .prop('checked', true)
+      .trigger('change');
+}
+
+/**
+ * Switches the schema of {@link globalTd} to the given type.
```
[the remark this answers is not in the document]
juliexxia: Added some more. Let me know if it's better.

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+  const colorFunction = getColorFunction();
+  const values =
+      colorFunction['columns'][$('#discrete-property-picker').val()]['values'];
+  const asColorPickers = [];
+  const colors = getColorFunction()['colors'];
+  for (const value of values) {
+    const li = $(document.createElement('li'));
+    li.append($(document.createElement('label')).text(value + ': '));
+    const colorPicker =
+        createColorPicker()
```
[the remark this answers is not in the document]
juliexxia: Done - not sure if in the best way.

docs/import/add_disaster.html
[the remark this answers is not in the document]
janakdr: Looks good! I would even be ok with the entire row being highlighted. Up to you.

docs/import/color_function_util.js
```diff
@@ -1,8 +1,162 @@
...
+// color function info is currently displayed.
+let globalTd;
+
+/**
+ * Fills out the #single #continuous and #discrete divs with the relevant
+ * DOM elements with attached on change handlers.
+ */
+function populateColorFunctions() {
+  const colorFunctionDiv = $('#color-fxn-editor');
+  colorFunctionDiv.prepend(createRadioFor(ColorStyle.SINGLE));
```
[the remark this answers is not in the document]
janakdr: Especially given the testing discussion we had, I'm good with keeping it in JS

docs/import/color_function_util.js
```diff
@@ -1,8 +1,165 @@
...
+function setColor(picker) {
+  getColorFunction()['color'] = picker.val();
+  updateLayersInFirestore();
+  populateColorTd();
+}
+
+/**
+ * Refreshes the current color cell.
+ */
+function populateColorTd() {
```
janakdr: As noted offline, it seems like this should be called on every single "pick", including changing the relevant property (in setProperty)? I wonder if the Firestore update should just be wrapped in a function that also updates the td. That way they'll never get out of sync.

docs/import/color_function_util.js
```diff
@@ -1,8 +1,165 @@
...
+ * @param {JQuery<HTMLElement>} picker
+ */
+function setColor(picker) {
+  getColorFunction()['color'] = picker.val();
+  updateLayersInFirestore();
+  populateColorTd();
+}
+
+/**
+ * Refreshes the current color cell.
```
janakdr: nit: can one-line

docs/import/color_function_util.js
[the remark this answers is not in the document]
janakdr: As discussed offline, TODO and remove in follow-up?

docs/import/color_function_util.js
```diff
@@ -1,8 +1,162 @@
...
+}
+
+/**
+ * Updates an individual color choice for a single value in the discrete schema.
+ * @param {JQuery<HTMLElement>} picker
+ */
+function setDiscreteColor(picker) {
+  const colorFunction = getColorFunction();
+  const propertyValue = picker.data('value');
+  if (!colorFunction['colors']) {
```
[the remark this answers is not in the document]
janakdr: TODO as discussed offline?

docs/import/add_layer.js
```diff
@@ -0,0 +1,33 @@
+import {convertEeObjectToPromise} from '../map_util.js';
+
+import {getCurrentLayers, updateLayersInFirestore} from './add_disaster_util.js';
+
+export {processNewFeatureLayer};
+
+const layerIndex = 4;
+
+/**
+ * One-off function for processing a feature collection typed layer and putting
```
janakdr: nit: "feature-collection-typed" :) or something like that

docs/import/add_disaster_startup.js
```diff
@@ -22,6 +23,8 @@ const authenticator = new Authenticator(
     });
 authenticator.start();
 
+$(populateColorFunctions());
```
janakdr: I think you need to do $(populateColorFunctions), not what you have. What you have does $() around the result of populateColorFunctions(), which is executed immediately. It seems like Javascript modules are executed after page load, which I can't find documented anywhere, but confirmed experimentally. So $() is overkill, but probably good practice.

docs/import/color_function_util.js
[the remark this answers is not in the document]
janakdr: As discussed offline, I'm in favor of standardizing the sync with Firestore so that the page view always corresponds to what's in Firestore. Add TODOs here and elsewhere to get rid of these guards once data is eagerly populated?

docs/import/color_function_util.js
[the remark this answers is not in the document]
juliexxia: I moved this check down a level inside populateDiscreteColorPickers and I think it should stay there. If the 'field' field isn't set to anything, it doesn't make sense to create any value color pickers right?

docs/import/color_function_util.js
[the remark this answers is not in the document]
juliexxia: Same question as below - is the TODO just to create the 'colors' map on layer creation?

docs/import/color_function_util.js
juliexxia: Done.

docs/import/add_disaster_startup.js
juliexxia: Done.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
+import * as addDisasterUtil from '../../../docs/import/add_disaster_util.js';
+import {populateColorFunctions, withColor} from '../../../docs/import/color_function_util.js';
+import {createTrs, setDisasterAndLayers} from '../../support/import_test_util.js';
+import {addFirebaseHooks, loadScriptsBeforeForUnitTests} from '../../support/script_loader.js';
+
+describe('Unit tests for add_disaster page', () => {
```
janakdr: nit: different text here?

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+    colorFunctionEditor =
+        $(document.createElement('div')).prop('id', 'color-fxn-editor').hide();
+    colorFunctionEditor.append(
+        makeTypeDiv('single'), makeTypeDiv('continuous'),
+        makeTypeDiv('discrete'));
+    $(document.body).append(colorFunctionEditor);
+
+    populateColorFunctions();
+
+    /**
```
janakdr: nit: just a style preference, but pull this out to the top level if it doesn't need to be nested?

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+
+    const td = withColor($(document.createElement('td')), layer, property);
+    const row = createTrs(1);
+    row[0].append(td);
+
+    expect(colorFunctionEditor.is(':visible')).to.be.false;
+
+    td.trigger('click');
+    expect(colorFunctionEditor.is(':visible')).to.be.true;
+    expect(writeToFirebaseStub).to.be.calledOnce;
```
janakdr: we write on color dialog display, even if there's no actual change? Seems like it would be good to avoid that, but it might be hard with my suggestion to set data unconditionally. Will comment below.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+    expect(getColorFunction()['color']).to.equal('yellow');
+
+    // update color
+    $('#single-color-picker').val('red').trigger('change');
+    expect(writeToFirebaseStub).to.be.calledTwice;
```
janakdr: I was wrong about stubs not having reset. I think .resetHistory() works, if you want to try that?

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+    $('#CONTINUOUS-radio').trigger('change');
+    expect(writeToFirebaseStub).to.be.calledThrice;
+    const continuousPropertyPicker = $('#continuous-property-picker');
+    expect(getColorFunction()['current-style']).to.equal(0);
+    expect(continuousPropertyPicker.val()).to.be.null;
+
+    // update field
+    continuousPropertyPicker.val('wings').trigger('change');
+    expect(writeToFirebaseStub).to.be.callCount(4);
+    expect(getColorFunction()['field']).to.equal('wings');
```
janakdr: Can you assert on the color here? Just so the expectation is clear.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+
+    // switch to discrete
+    $('#DISCRETE-radio').trigger('change');
+    expect(writeToFirebaseStub).to.be.callCount(5);
```
janakdr: super-nit: does "to.have" work instead?

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -0,0 +1,133 @@
...
+    const discretePropertyPicker = $('#discrete-property-picker');
+    expect(getColorFunction()['current-style']).to.equal(1);
+    expect(td.children().length).to.equal(0);
+    expect(discretePropertyPicker.val()).to.equal('wings');
+    const discreteColorPickerList = $('#discrete-color-pickers');
```
janakdr: should the field be wings at this point?

docs/import/color_function_util.js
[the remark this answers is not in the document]
janakdr: Yup!

docs/import/color_function_util.js
```diff
@@ -278,9 +290,13 @@ const discreteColorPickerDataKey = 'value';
...
   const pickerList = $('#discrete-color-pickers').empty();
   const colorFunction = getColorFunction();
+  if (!colorFunction['field']) {
+    return;
+  }
   const values =
       colorFunction['columns'][$('#discrete-property-picker').val()]['values'];
   const asColorPickers = [];
+  // TODO: remove, always have this around at this point.
   const colors = getColorFunction()['colors'];
```
janakdr: nit: use colorFunction instead of getting again?

docs/import/color_function_util.js
```diff
@@ -25,22 +191,149 @@ function withColor(td, layer, property, index) {
...
+  }
+  colorFunctionDiv.show();
+  if (td === globalTd) {
+    $(globalTd).addClass('selected');
+    return;
+  }
+  $(globalTd).removeClass('selected');
+  globalTd = td;
+  $(globalTd).addClass('selected');
+  $('#' + colorStyleTypeStrings.get(type) + '-radio')
```
janakdr: So we're assuming there is always a default type already set? Maybe we can make that SINGLE_COLOR on layer creation?

docs/import/color_function_util.js
```diff
@@ -25,22 +191,149 @@ function withColor(td, layer, property, index) {
...
+      .prop('checked', true)
+      .trigger('change');
```
janakdr: This trigger ends up writing to Firestore, right? What do you think of a method displaySchema(type) that has all of setSchema except for the last two lines. Then switchSchema can call it, and here, instead of triggering a change, you can call displaySchema. That way there's no writing to Firestore on clicking into the color.

docs/import/color_function_util.js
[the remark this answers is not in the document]
janakdr: Sounds good.

janakdr:
> In docs/import/color_function_util.js
> <https://github.com/givedirectly/Google-Partnership/pull/238#discussion_r350513220>
> :
>
> > + * color schemas.
> + * @param {JQuery<HTMLElement>} picker
> + */
> +function setProperty(picker) {
> +  getColorFunction()['field'] = picker.val();
> +  updateLayersInFirestore();
> +}
> +
> +/**
> + * Updates an individual color choice for a single value in the discrete schema.
> + * @param {JQuery<HTMLElement>} picker
> + */
> +function setDiscreteColor(picker) {
> +  const colorFunction = getColorFunction();
> +  const propertyValue = picker.data('value');
> +  if (!colorFunction['colors']) {
>
> Same question as below - is the TODO just to create the 'colors' map on
> layer creation?
>
Yes, I think so.

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: Done - with assertion that nothing is called on display

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: Yeah - this test is probably too big. But I think it covers most of the complexity.

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: clearing history

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: yep, added an expect

docs/import/color_function_util.js
juliexxia: Yep, that's what I was thinking. Maybe they also pick a starting color but I guess that doesn't need to happen.

docs/import/color_function_util.js
juliexxia: Done.

docs/import/color_function_util.js
```diff
@@ -25,22 +179,136 @@ function withColor(td, layer, property, index) {
...
+}
+
+/**
+ * Creates the set of color boxes for a discrete schema.
+ * @param {Object} colorFunction
+ * @param {JQuery<HTMLElement>} td cell in which to create the boxes.
+ */
+function createColorBoxesForDiscrete(colorFunction, td) {
+  const colorObject = colorFunction['colors'];
+  // coming from a place that has never had discrete before
```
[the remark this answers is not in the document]
juliexxia: Discussed offline - will do in followup

docs/import/color_function_util.js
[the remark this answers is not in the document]
juliexxia: Done.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -3,13 +3,14 @@ import {populateColorFunctions, withColor} from '../../../docs/import/color_func
 import {createTrs, setDisasterAndLayers} from '../../support/import_test_util.js';
 import {addFirebaseHooks, loadScriptsBeforeForUnitTests} from '../../support/script_loader.js';
 
-describe('Unit tests for add_disaster page', () => {
+const property = 'color-function';
```
janakdr: nit again :) What I said about declaring as globals for functions doesn't apply to variables: https://google.github.io/styleguide/jsguide.html#features-declared-when-needed. The idea is that functions capture their scopes going up, so you want to make that scope as small as possible (by making them global), while variables are captured by all the scopes going down, so you want to make that as small as possible by making them local.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -115,19 +108,38 @@ describe('Unit tests for add_disaster page', () => {
...
-  /**
-   * Gets the current color function.
-   * @return {Object}
-   */
-  function getColorFunction() {
-    return addDisasterUtil.getCurrentLayers()[0][property];
-  }
 });
+
+/** Assert we wrote to firebase once and clear stub history. */
```
janakdr: nit: Asserts

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: These need to be up here because the functions need to be able to access them. Seems like competing interests!

janakdr: Ah, sorry for being unclear. In my initial nit I was only referring to the "makeTypeDiv" function, since I didn't see it use any variables from the immediate scope, not the other functions you had. The ones that use variables that are mutated in the describe can, of course, live inside the describe!

PR #239, opened 2019-11-23
Combine 'single-color' and 'base-color' fields in firestore

juliexxia:
We have an explicit current schema field now so we don't need these to be separate. This also means if a user is switching between continuous and single the property gets reused across the different schemas.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,7 +26,7 @@ We have three types of coloring schemes: continuous, discrete, and single color.
 
 **continuous:**
 * current-style `{enum}`: set to 0
-* color `{string}`: from known colors
+* color `{string}`: from set of known colors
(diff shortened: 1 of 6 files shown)


juliexxia:
Reminder to self - purge all instances of 'single-color' and 'base-color' in firestore when this goes in.

docs/README.md
```diff
@@ -40,7 +40,7 @@ We have three types of coloring schemes: continuous, discrete, and single color.
 
 **single-color:**
 * current-style `{enum}`: set to 2
-* single-color `{string}`: from set of known colors
+* color `{string}`: from set of known colors
```
janakdr: Nit: not new, but keep wording consistent with above?

PR #240, opened 2019-11-24
Refactor import_data to pull data from Firestore, not hard-coded resources

janakdr:
Also change import_data.js in the following ways:
1. Allow multiple states to be entered, combining them all into one big score asset.
2. Tolerate the possible lack of a damage asset, instead using user-supplied map bounds. I think that, given the [ease of getting lat-long coordinates from Google Maps](https://support.google.com/maps/answer/18539), we can ask users to enter them on add_disaster.html before the damage asset is available.
3. Use an "envelope" around the damage bounds for clipping other assets, so that we don't exclude block groups that are near damage but don't have any damage themselves. Maps page is a bit slower, but still ok.
4. Change the calculation of building count per block group, just for fun, since I hope that EarthEngine optimizes joins more than the hack we were doing before.
5. Treat no buildings as no damage, to avoid bad data giving us weird results.

Other main code change is to standardize the pages that need to do EarthEngine and Firebase authentication, by providing a utility method in Authenticator that does what all of them want. This has the side effect of making all those pages use the EE cloud API, which is probably a good thing anyway.

It's not clear to me that we'll actually want the user to have to choose all of the values that import_data is now reading. We can always inject them by setting them in Firestore when the disaster is created. Then if later we decide that the user should choose them, we can add a field in add_disaster.html.

I've entered all the data for Harvey into Firestore manually, but currently this won't work with a new disaster. We'll have to modify add_disaster.html to either inject or have the user set values.

I made the naming convention in Firestore use '\_' rather than '-' because it makes Javascript property names and constructing test objects so much easier to write and read. If the inconsistency disturbs you, we can migrate the other fields in our database to use '\_' (or camelCase, which might make our linter happier sometimes).


janakdr:
@ruthtalbot FYI

docs/import/import_data.js
```diff
@@ -293,16 +305,20 @@ function run(disasterData) {
...
+const damageBuffer = 1000;
+
 /**
  * Calculates damage if there is a damage asset, or simply writes the
  * user-provided bounds to Firestore if not.
  * @param {Object} assetData
- * @return {{damage: ?ee.FeatureCollection, mapBoundsRectangle:
- *     ee.Geometry.Rectangle}|{damage: null, mapBoundsRectangle: null}} Returns
- *     the damage asset (if present) and the rectangle bounding the damage, or
+ * @return {{damage: ?ee.FeatureCollection, damageEnvelope:
```
janakdr: Don't know what's going on here.

janakdr:
Updated PR comment to reflect current state of world.

docs/import/import_data.js
```diff
@@ -140,132 +135,327 @@ function stringifyGeoid(feature) {
...
+        .then((bounds) => centerStatusSpan.innerText = 'Found bounds ' + bounds)
+        .catch((err) => centerStatusSpan.innerText = err);
+    return {damage, damageEnvelope: damage.geometry().buffer(damageBuffer)};
+  }
+  // TODO(janakr): in the no-damage case, we're storing a rectangle, but
+  //  experiments show that, at least for Harvey, the page is very slow when we
+  //  load the entire rectangle around the damage. Maybe allow users to select a
+  //  polygon so they can draw a tighter area?
+  centerStatusLabel.innerText = 'Storing bounds of map: ';
+  const damageSw = assetData['map_bounds_sw'];
```
juliexxia: maybe this should also go under map-bounds not asset-data?

docs/import/import_data.js
```diff
@@ -140,132 +135,327 @@ function stringifyGeoid(feature) {
...
+ * This method will go away or be greatly changed if we're using CrowdAI data
+ * instead of previously computed building data.
+ * @param {ee.Geometry.Polygon} damageEnvelope Area we are concerned with
+ * @param {string} buildingPath location of buildings asset in EE
+ * @param {string} state
+ * @param {ee.FeatureCollection} stateGroups Collection with block groups
+ * @return {ee.Dictionary} Number of buildings per block group
+ */
+function computeBuildingsHisto(
+    damageEnvelope, buildingPath, state, stateGroups) {
```
juliexxia: unused param state? Don't know why intellij isn't upset about this.

docs/import/import_data.js
```diff
@@ -140,132 +135,327 @@ function stringifyGeoid(feature) {
...
+  const buildings =
+      ee.FeatureCollection(buildingPath).filterBounds(damageEnvelope);
   const withBlockGroup =
-      buildings.map((building) => attachBlockGroups(building, blockGroups));
-  const buildingsHisto =
-      ee.Dictionary(withBlockGroup.aggregate_histogram(geoidTag));
-  // process final feature collection
-  const data = joinedSnapIncomeSVI.map(
-      (feature) => countDamageAndBuildings(feature, buildingsHisto));
+      ee.Join.saveFirst('bg')
```
juliexxia: make 'bg' string a constant?

docs/resources.js
```diff
@@ -24,58 +15,11 @@ function getDisaster() {
 
 /** @return {string} EE asset path for score asset for the current disaster */
 function getScoreAsset() {
-  return gdEePathPrefix + getDisaster() + '/data-ms-as-nod';
+  // TODO(janakr): Modify map to handle no-damage scenario more intuitively.
+  //  Should make damage threshold, weight 0, probably hide those toggles
+  //  completely. Other changes?
+  return gdEePathPrefix + getDisaster() + '/data-ms-as-tot';
```
juliexxia: Maybe make this path a constant so it's not repeated in resources.js?

docs/import/import_data.js
```diff
@@ -140,132 +135,327 @@ function stringifyGeoid(feature) {
...
+    allStatesProcessing = allStatesProcessing.merge(processing);
+  }
+
+  const assetName = 'data-ms-as-tot';
+  const scoreAssetPath = gdEePathPrefix + getDisaster() + '/' + assetName;
+  try {
+    ee.data.deleteAsset(scoreAssetPath);
+  } catch (err) {
+    if (err.message !== 'Asset not found.') {
+      throw err;
```
juliexxia: Do we want to throw an error when there isn't an old asset to delete? What about the first time running this script?

docs/import/center.js
```diff
@@ -1,55 +1,47 @@
...
-  const outerBounds = ee.List([
-    damageWithCoords.aggregate_min('lat'),
-    damageWithCoords.aggregate_min('lng'),
-    damageWithCoords.aggregate_max('lat'),
-    damageWithCoords.aggregate_max('lng'),
-  ]);
-  return Promise
-      .all([firebaseAuthPromise, convertEeObjectToPromise(outerBounds)])
-      .then((results) => saveBounds(results[1]));
+function storeCenter(features) {
```
juliexxia: rename? Doesn't do anything with centers anymore.

cypress/integration/unit_tests/import_data_test.js
```diff
@@ -0,0 +1,216 @@
...
+  let exportStub;
+  beforeEach(() => {
+    // Create a pretty trivial world: 2 block groups, each a 1x2 vertical
+    // stripes. Under the covers, we scale all dimensions down because
+    // production code creates an "envelope" 1 km wide around damage, and that
+    // envelope is assumed to fully contain any block group that has any damage.
+    const tigerBlockGroups = ee.FeatureCollection(
+        [makeCensusBlockGroup(0), makeCensusBlockGroup(1)]);
+    // Three damage points, one of them outside the blocks, just for fun.
+    // Relevant damage points are in SW and SE blocks.
```
juliexxia: what does SW and SE blocks mean?

cypress/integration/unit_tests/import_data_test.js
```diff
@@ -0,0 +1,216 @@
...
+          NY: incomeData,
+        },
+        income_key: 'testIncomeKey',
+        building_asset_paths: {
+          NY: buildingsCollection,
+        },
+      },
+    };
+  });
+  it('Basic test', () => {
```
juliexxia: basic but maybe add a test for when run doesn't return true?

cypress/integration/unit_tests/import_data_test.js
janakdr: Mostly left over from when I had blocks. Rephrased.

docs/import/import_data.js
```diff
@@ -1,65 +1,59 @@
...
- * the source assets are already uploaded. Uploading a .csv or .shp can be done
- * with something like the command line:
- * `earthengine upload table --asset_id users/janak/census_building_data \
- *      gs://noaa-michael-data/ACS_16_5YR_B25024_with_ann.csv`
- * (assuming the file has already been uploaded into Google Cloud Storage).
+ * TODO: This should be moved to a user-readable doc and expanded.
+ * Processes all user-entered data to construct the "score" asset in EarthEngine
+ * and write map bounds for the disaster to Firestore. Mandatory data
+ * (state-level data needs to be present for every affected state):
+ * 1. National Census block shapes, normally provided by EarthEngine.
```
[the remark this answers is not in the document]
janakdr: Thanks, rewrote.

docs/import/import_data.js
```diff
@@ -1,65 +1,59 @@
...
- * 7) update the {@code disaster} constant in ./resources.js
- * 8) visit http://localhost:8080/import_data.html
- * 9) make the new asset readable by all in code
- * editor.
+ * 3. download crowd ai damage data .shp file
+ * 4. upload results of (1) and (2) to EarthEngine via the code editor
+ * 5. visit http://localhost:8080/import/add_disaster.html and fill out all
+ * required fields
+ * 6. visit http://localhost:8080/import/import_data.html and submit the page.
+ * 7. make the created asset readable by all in the code editor.
```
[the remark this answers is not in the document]
janakdr: I don't think so, you can't set ACLs on an export task, and we can't set the ACL of the asset until the export task has finished, which could be hours. The page could keep polling to see if the job was finished, and set the ACL when it did, but it just may not work out.

docs/import/import_data.js
```diff
@@ -68,17 +62,31 @@ function countDamageAndBuildings(feature, buildings) {
...
-          .set(damageTag, ratioBuildingsDamaged));
+          .set(buildingCountTag, totalBuildings);
+  if (damage) {
+    const damagedBuildings =
+        ee.FeatureCollection(damage).filterBounds(geometry).size();
+    properties = properties.set(
+        damageTag,
+        // If no buildings, this is probably spurious. Don't give any damage.
+        ee.Algorithms.If(
+            totalBuildings, ee.Number(damagedBuildings).divide(totalBuildings),
```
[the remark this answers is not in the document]
janakdr: Yup. But I think a no-building area is likely to be unimportant to GD anyway. If CrowdAI has a false positive in the middle of a forest, I think we want to filter it out.

docs/import/import_data.js
```diff
@@ -140,132 +135,357 @@ function stringifyGeoid(feature) {
...
-  // since we're only using block groups from damaged areas, we have buildings
-  // that won't intersect any block groups.
-  const geoid = ee.Algorithms.If(
-      filtered.size().gt(ee.Number(0)), filtered.first().get(tigerGeoidKey),
-      ee.String('PLACEHOLDER GEOID VALUE'));
-  return building.set(geoidTag, geoid);
+ * Performs operation of processing inputs and creating output asset.
+ * @param {Object} disasterData Data for current disaster coming from Firestore
+ * @return {boolean} Whether in-thread operations succeeded
+ * */
```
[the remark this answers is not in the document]
janakdr: eslint, you had one job

docs/import/import_data.js
```diff
@@ -140,132 +135,357 @@ function stringifyGeoid(feature) {
...
+function run(disasterData) {
+  $('.compute-status').html('');
+  const states = disasterData['states'];
+  if (!states) {
+    return missingAssetError('affected states');
```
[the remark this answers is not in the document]
janakdr: I think that the add_disaster page should warn/error if you're missing any fields, and only point you here once they're all ready. This is just an extra check, I think user friendliness should go there. Sound ok?

docs/import/import_data.js
janakdr: We're throwing an error only if the error is *not* "Asset not found". Rearranged to make clearer.

docs/import/import_data.js
```diff
@@ -140,132 +135,357 @@ function stringifyGeoid(feature) {
...
-              bounds[0] + ', ' + bounds[1] + '), (' + bounds[2] + ', ' +
-              bounds[3] + ')')
-      .catch((err) => centerStatusSpan.innerText = err);
-  $('.compute-status').append(centerStatusLabel);
-  $('.compute-status').append(centerStatusSpan);
+  const firestoreError = (err) => centerStatusSpan.innerText +=
+      'Error writing bounds to Firestore: ' + err;
+  if (damagePath) {
+    const damage = ee.FeatureCollection(damagePath);
+    // Uncomment to test with a restricted damage set (14 block groups' worth).
```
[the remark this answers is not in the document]
janakdr: Yes, it's annoying to have to paste it back in every time you're debugging something.

docs/import/import_data.js
janakdr: My conception is that asset_data is about the raw ingredients that go into the final data that the map uses, while map-bounds is that final data. So I think some separation is good.

docs/import/import_data.js
janakdr: Thanks. Yeah, I've noticed that before about intellij with javascript and I don't understand it.

docs/import/import_data.js
```diff
@@ -12,22 +11,24 @@ export {countDamageAndBuildings, run};
...
- * 1. National Census block shapes, normally provided by EarthEngine.
- * 2. State-level Census SNAP data, at the block group level.
- * 3. State-level SVI data, at the Census tract level.
- * 4. State-level income data, at the block group level.
- * 5. State-level FeatureCollection of all buildings.
+ * and write map bounds for the disaster to Firestore. Mandatory state-level
+ * data (needs to be present for every affected state):
+ * 1. Census block group shapes (TIGER shapefile).
+ * 2. Census SNAP data, at the block group level.
+ * 3. SVI data, at the Census tract level.
```
juliexxia: nit - using both state-level and tract level here - standardize hyphenation and maybe phrase one differently?

cypress/integration/unit_tests/import_data_test.js
```diff
@@ -0,0 +1,222 @@
...
+          total_key: 'test_total_key',
+        },
+        svi_asset_paths: {
+          NY: sviData,
+        },
+        svi_key: 'test_svi_key',
+        income_asset_paths: {
+          NY: incomeData,
+        },
+        income_key: 'testIncomeKey',
```
juliexxia: nit: snake case?

docs/import/center.js
```diff
@@ -1,55 +1,47 @@
...
-    damageWithCoords.aggregate_min('lat'),
-    damageWithCoords.aggregate_min('lng'),
-    damageWithCoords.aggregate_max('lat'),
-    damageWithCoords.aggregate_max('lng'),
-  ]);
-  return Promise
-      .all([firebaseAuthPromise, convertEeObjectToPromise(outerBounds)])
-      .then((results) => saveBounds(results[1]));
+function computeAndSaveBounds(features) {
+  return convertEeObjectToPromise(features.geometry().bounds())
```
juliexxia: is features.geometry a no no?

docs/resources.js
juliexxia: Did this change?

cypress/integration/unit_tests/import_data_test.js
janakdr: keys can be any strings?

docs/import/center.js
janakdr: This is way faster than the previous method. I agree we were told not to use FeatureCollection.geometry(), but the reason was that it's a huge bag of geometries. However, once you do .bounds(), it's just a rectangle. So not a lot of data.

docs/import/import_data.js
janakdr: I agree this grammar rule is confusing! "state-level" is hyphenated because it's an adjective, modifying "data". In "Census tract level", "level" is the noun, and "Census tract" is a compound adjective, modifying it. I could hyphenate it like "Census-tract level", but the rule is that if the compound adjective is common enough, or if the individual components aren't applicable adjectives themselves, you don't need to hyphenate it. I would say that applies to both "Census tract level" and "block group level". 

However, if the wording is hard to comprehend, happy to rephrase. Can you say if something in particular is confusing?

docs/import/import_data.js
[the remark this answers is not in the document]
janakdr: well, the "0" branch will be triggered if totalBuildings evaluates to false. EarthEngine says in their docs:

> The condition that determines which result is returned. If this is not a boolean, it is interpreted as a boolean by the following rules:
> - Numbers that are equal to 0 or a NaN are false.
> - Empty strings, lists and dictionaries are false.
> - Null is false.
> - Everything else is true.

Since totalBuildings is either 0, or the value of a histogram, it is a number. So unless we somehow got NaN from the histogram, totalBuildings being false indicates we found no buildings in that block group. Is that what you were asking? Sorry if I misunderstood.

docs/import/import_data.js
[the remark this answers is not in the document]
janakdr: I think having a button to actually kick off the process makes sense, especially given that it will delete the previous version of the asset.

docs/resources.js
janakdr: Yes, import_data is now just calling this to get the correct path.

PR #242, opened 2019-11-25
Use relative path to load navbar

juliexxia:
This should fix being able to see the navbar on our github pages served site


janakdr:
See also https://github.com/givedirectly/Google-Partnership/pull/243

PR #244, opened 2019-11-26
Fix relative navbar loading issues

janakdr:
And pull on some rather annoying strings: make more of our pages work with our test setup (of Firestore and EE tokens passed via cookies) by pushing the test branch into the new Authenticator static method, and change eslint a bit, because it doesn't like the "import.meta" tokens.


janakdr:
@RamV13 FYI

PR #246, opened 2019-11-26
Fix import data test

janakdr:
We were kicking off the Firestore fetch without waiting for anything. ~I still don't understand how that fetch could see data from a different test case (because we wipe data between tests), but~ 
 What we were doing before was clearly wrong on two levels: we weren't waiting for the write to complete, and we were doing the read in program order, not Cypress order. Fixed both of those, with not too bad a hack on the first side, I hope.

Also, we weren't actually doing any proper setup in import_data_test for Firebase, which explains why the data was sticking around. Fixed that. We didn't need to do proper setup because I had apparently made the database global read-write during testing and forgotten to restrict it. I've tested with the restriction (it works) and undone it for now. Will restrict after this PR is in.

Also fixed an off-by-one error in the task description for import_data.js's EE export. ~, and temporarily use the old score asset since I was an idiot and deleted it while testing this page.~

Also extend polygon_draw_test timeout a bit.

### DIFF
--- a/cypress/support/firestore_map_bounds.js
+++ b/cypress/support/firestore_map_bounds.js
@@ -9,7 +9,8 @@ export {assertFirestoreMapBounds, expectLatLngBoundsWithin};
  *     expectedLatLngBounds
  */
 function assertFirestoreMapBounds(expectedLatLngBounds) {
-  cy.wrap(readDisasterDocument()).then((doc) => {
+  // Make sure we don't call readDisasterDocument until Cypress is ready.
+  cy.wrap(null).then(readDisasterDocument).then((doc) => {
     // Expect that result retrieved from Firestore is correct.
     const mapBounds = doc.data()['map-bounds'];
     expectLatLngBoundsWithin(
--- a/cypress/support/script_loader.js
+++ b/cypress/support/script_loader.js
@@ -4,7 +4,7 @@ import {cypressTestCookieName, earthEngineTestTokenCookieName, firebaseTestToken
 export {
   addFirebaseHooks,
   initFirebaseForUnitTest,
-  loadScriptsBeforeForUnitTests
+  loadScriptsBeforeForUnitTests,
 };
 
 /**
--- a/cypress/support/script_loader.js
+++ b/cypress/support/script_loader.js
@@ -1,7 +1,7 @@
 import {CLIENT_ID, getFirebaseConfig} from '../../docs/authenticate';
 import {cypressTestCookieName, earthEngineTestTokenCookieName, firebaseTestTokenCookieName} from '../../docs/in_test_util';
 
-export {addFirebaseHooks, loadScriptsBeforeForUnitTests};
+export {addFirebaseHooks, loadScriptsBeforeForUnitTests, initFirebaseForUnitTest};
 
 /**
  * Scripts that unit tests may want to load. Values have script and callback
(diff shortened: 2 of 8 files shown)


janakdr:
@RamV13 @ruthtalbot I think Julie might be offline for a while, so if either of you wants to review, please do!

cypress/support/firestore_map_bounds.js
```diff
@@ -9,7 +9,8 @@ export {assertFirestoreMapBounds, expectLatLngBoundsWithin};
  *     expectedLatLngBounds
  */
 function assertFirestoreMapBounds(expectedLatLngBounds) {
-  cy.wrap(readDisasterDocument()).then((doc) => {
+  // Make sure we don't call readDisasterDocument until Cypress is ready.
+  cy.wrap(null).then(readDisasterDocument).then((doc) => {
```
juliexxia: cypress was executing the code inside cy.wrap before cypress was ready?

docs/resources.js
```diff
@@ -18,7 +18,9 @@ function getScoreAsset() {
   // TODO(janakr): Modify map to handle no-damage scenario more intuitively.
   //  Should make damage threshold, weight 0, probably hide those toggles
   //  completely. Other changes?
-  return gdEePathPrefix + getDisaster() + '/data-ms-as-tot';
+  // TODO(janakr): switch back to -as-tot when regenerated, and make a backup
+  //  so you don't lose it again.
+  return gdEePathPrefix + getDisaster() + '/data-ms-as-nod';
```
juliexxia: switch back?

cypress/support/firestore_map_bounds.js
janakdr: Yeah, remember that Cypress has no superpowers. The Javascript interpreter still executes the code, and Cypress just makes sure that its internal commands don't execute until certain conditions are met. Thus, cy.wrap(blah()) will execute blah() and pass the result to cy.wrap. Cypress can then make sure a later cy.get() doesn't do anything until the blah result completes (because cy.get() itself is just putting a command on the Cypress queue to execute later), but it can't control blah() executing.

cypress/support/script_loader.js
```diff
@@ -124,10 +128,9 @@ function loadScriptsBeforeForUnitTests(...scriptKeys) {
     });
   }
   if (usesFirebase) {
-    // Currently no unit test actually writes to Firestore, they just use the
-    // library. A unit test that really writes to Firestore will be responsible
-    // for calling the necessary functions (addFirebaseHooks and logging in with
-    // the custom token, similar to what is done in prod).
+    // Some unit tests just use the Firebase library, so that's all we prepare
+    // for here. A unit test that really writes to Firestore has to call
+    // initFirebaseForUnitTest.
```
juliexxia: why not always do initFirebaseForUnitTest? Are there a lot of tests where we use the script and don't actually read/write?

janakdr: Currently just unit_tests/polygon_draw_test. I thought about it, but this keeps this PR minimal in terms of functional change, and this does add some latency to each test case. I think that in a follow-up I'll have unit_tests/polygon_draw_test actually assert on the Firestore write contents. That might also allow us to delete some parts of integration_tests/polygon_draw_test with a clearer conscience

juliexxia: sgtm!

PR #247, opened 2019-11-27
Partially deflake polygon_draw_test

janakdr:
This is actually a bug in production, though a very hard one to hit. We were assuming that by the time the calculated data for a polygon was ready, the popup's html was created. However, the popup's html is normally created in onAdd, called by the Google Map, which happens in a separate thread from popup creation (verified with console debug statements). Thus, it is possible for the calculated data task to complete and try update calculated data before the div is present. Tolerate that situation: when the div is created, it will be initialized to the proper value in createPopupHtml() anyway.


juliexxia:
Great find

PR #248, opened 2019-11-27
Add ability to add layer

juliexxia:
Add ability to add a layer from a picker of disaster assets and state assets. The layer automatically goes on top. We preprocess all the property values so we know mins, maxes, and if <= 25 values, then all the values. If a property has > 25 values we don't allow users to color that property discretely. 

Also stop allowing user to specify opacity. Use a standard opacity instead and Alex can tell us if it's not good enough.

TODO: 
- stop checking if colors map exists since we're always creating now
- get rid of opacity in firebase if Alex doesn't care.


docs/import/add_layer.js
```diff
@@ -1,32 +1,73 @@
...
+            featureCollection.aggregate_histogram(property), ee.Dictionary());
+        return ee.Dictionary.fromLists(
+            ['max', 'min', 'values'], [max, min, values]);
       });
+      return convertEeObjectToPromise(
+                 ee.Dictionary.fromLists(properties, stats))
+          .then((columns) => {
+            for (const property of Object.keys(columns)) {
+              columns[property]['values'] =
+                  Object.keys(columns[property]['values']);
```
janakdr: Why not do this in the EE function, so that we don't have to download the dictionary, we can just download a list?

docs/import/add_layer.js
```diff
@@ -1,32 +1,73 @@
...
+                .lte(ee.Number(25)),
+            featureCollection.aggregate_histogram(property), ee.Dictionary());
+        return ee.Dictionary.fromLists(
+            ['max', 'min', 'values'], [max, min, values]);
       });
+      return convertEeObjectToPromise(
+                 ee.Dictionary.fromLists(properties, stats))
+          .then((columns) => {
+            for (const property of Object.keys(columns)) {
+              columns[property]['values'] =
```
janakdr: Probably the friendly thing to do to alphabetize these values. Leave a TODO at least?

docs/import/add_layer.js
```diff
@@ -1,32 +1,73 @@
...
+    case LayerType.FEATURE_COLLECTION:
+      const featureCollection = ee.FeatureCollection(asset);
+      const properties = featureCollection.first().propertyNames();
+      const stats = properties.map((property) => {
+        const max = featureCollection.aggregate_max(property);
+        const min = featureCollection.aggregate_min(property);
+        const values = ee.Algorithms.If(
+            ee.Number(featureCollection.aggregate_count_distinct(property))
+                .lte(ee.Number(25)),
+            featureCollection.aggregate_histogram(property), ee.Dictionary());
```
janakdr: Should we filter out columns that have only one unique value? Or at least not let them be selected for the discrete coloring scheme?

docs/import/add_disaster.js
```diff
@@ -476,63 +514,101 @@ function getAssetsFromEe(states) {
...
+/**
+ * Create an asset picker for the given disaster.
+ * @param {string} disaster
+ */
+function createDisasterAssetPicker(disaster) {
+  createAssetPickers([disaster], disasterAssets, $('#disaster-asset-picker'));
+}
+/**
+ * Given either states or disasters, displays their assets in pickers. Right
+ * now, selecting on those pickers doesn't actually do anything.
```
janakdr: this isn't fully true anymore, right?

docs/import/add_disaster.js
```diff
@@ -225,67 +230,87 @@ function populateLayersTable() {
...
+  row.append(withType(createTd(), layer, 'asset-type'));
+  // display on load
+  row.append(withCheckbox(createTd(), layer, 'display-on-load'));
+  // color
+  row.append(withColor(createTd(), layer, 'color-function', index));
+  return row;
 }
 
 /**
  * Populates the state asset pickers with all known earth engine assets for
```
janakdr: nit: update docs?

docs/import/add_disaster.js
```diff
@@ -327,6 +352,7 @@ function initializeScoreSelectors(states) {
  * @param {Array<string>} assets array of assets for add to dropdown
  * @param {string} row The asset type/row to put the dropdown in.
  * @param {string} state The state the assets are in.
+ * @return {JQuery<HTMLSelectElement>}
  */
 function createAssetDropdown(assets, row, state) {
```
janakdr: I think this function is broken now, because it's iterating up to assets.length which is now undefined. May not matter that much, since I have to modify this function a lot, but might be nice to keep it un-broken. Also, for score assets, everything will be a feature collection. I wonder if we could thread the needle in terms of display by leaving out the FEATURE_COLLECTION modifier in general, assuming it's the default? Maybe not worth it.

docs/firebase_layers.js
```diff
@@ -50,22 +50,22 @@ function createStyleFunction(colorFunctionProperties) {
...
  */
-function createContinuousFunction(field, opacity, minVal, maxVal, color) {
+function createContinuousFunction(field, minVal, maxVal, color) {
   const colorRgb = colorMap.get(color);
   const white = colorMap.get('white');
   return (feature) => {
     const value = feature['properties'][field];
     const rgba = [];
     for (let i = 0; i < 3; i++) {
+      // https://www.alanzucconi.com/2016/01/06/colour-interpolation/
```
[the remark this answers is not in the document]
janakdr: Sure. Note that the only difference between this formula and the one we had before was the scaling. The old one just had a bug where we weren't scaling down properly (dividing by 2 instead of the range).

docs/import/add_disaster.js
janakdr: Yeah, I think it won't be super-useful for them, but not a big deal

docs/import/add_layer.js
janakdr: If you try to add "path of the storm" feature collection, you get a whole bunch of columns, and no easy indication which is the one you should be using. Filtering out all the ones that are constant would make that a lot easier.

PR #249, opened 2019-12-01
Delete cleanup_acs.sh script: EE now does it for us

janakdr:
Tested that column names with '.' in them are accepted now: EE automatically removes the dots, same as our script did.

Also removed the documentation in import_data.js, since we're moving it to Google Docs (pasted the text into the Delphi documentation for cleanup).


juliexxia:
We still need to remove the second header row from census data. We could do that with the script or I guess we could do it in the earth engine join to tiger script. What do you think?

janakdr:
Definitely in the joining or just manually for now? We can tell users to remove the second row of a spreadsheet or a text file. Telling them to run a script is a higher bar.

juliexxia:
Sounds good to me, just wanted to keep that in our collective conscious (:

janakdr:
Added a comment in the instructions doc

PR #251, opened 2019-12-02
Add ability to delete a layer

juliexxia:


docs/import/add_disaster.js
```diff
@@ -78,6 +79,19 @@ function toggleDisaster(disaster) {
   return populateStateAssetPickers();
 }
 
+/**
+ * Reindex table rows in-between bounds (inclusive).
```
janakdr: nit: s/in-between/in between/

docs/import/add_disaster.js
```diff
@@ -96,11 +110,9 @@ function updateAfterSort(ui) {
   layers.splice(newRealIndex, 0, row);
 
   // Reindex all the layers.
```
janakdr: nit: not from this PR, but "all the layers" is misleading, I think

PR #253, opened 2019-12-02
Fix a bug where we weren't initializing a layer array on new disaster creation

juliexxia:
This was causing undefined errors later on


janakdr:
This is fixed in my pending PR. Mind waiting for that?

juliexxia:
Usurped by #255

PR #254, opened 2019-12-02
Fix navbar issues by using the replace trick

juliexxia:


docs/navbar.js
```diff
@@ -35,6 +39,16 @@ function loadNavbarWithPicker(firebaseAuthPromise) {
...
                 .load(
-                    '/disaster_picker.html',
+                    getUrlUnderDocs('disaster_picker.html'),
                     () => initializeDisasterPicker(firebaseAuthPromise)));
 }
+
+/**
+ * Get url of a file in or below our docs directory by using the path of this
+ * current script.
+ * @param {string} replaceWith
```
RamV13: I feel like a better name might just be `path` right?

juliexxia: oh yes, I renamed this method and forgot to appropriately rename the param. thanks!

PR #256, opened 2019-12-03
Have selectors display current score asset values, fix bug

janakdr:
Bugfix: unnest the asset data that's being created.
Functionality: have score selects display actual data, also show damage. We render them as soon as page loads, so the page jumps around less.

Still to do:
* Add selects for buildings, TIGER shapefiles.
* Write values back on change.
* Disable "kick off" button until have valid values.


docs/import/manage_disaster.html
```diff
@@ -104,12 +104,16 @@
       <div>
         <table id="asset-selection-table">
           <thead>
-          <tr id="score-asset-header-row">
+          <tr id="score-asset-header-row"><td>Score Assets</td>
```
juliexxia: super nit this is formatted kinda weird

docs/import/manage_disaster.js
```diff
@@ -675,41 +683,87 @@ function toggleState(known) {
...
 
+const scoreAssetTypes = [
+  ['poverty', ['snap_data', 'paths'], 'Poverty'],
+  ['income', ['income_asset_paths'], 'Income'],
+  ['svi', ['svi_asset_paths'], 'SVI'],
+];
+Object.freeze(scoreAssetTypes);
+
+const assetSelectionRowPrefix = 'asset-selection-row-';
+Object.freeze(assetSelectionRowPrefix);
```
juliexxia: is it worth freezing strings?

docs/import/manage_disaster.js
```diff
@@ -675,41 +683,87 @@ function toggleState(known) {
...
+    const propertyPath = scoreAssetType[1];
+    const row = $('#' + id);
+    removeAllButFirstFromRow(row);
     for (const state of states) {
       if (stateAssets.get(state)) {
+        const pathDictionary = getElementFromPath(propertyPath);
         const select =
-            createAssetDropdown(stateAssets.get(state), scoreAssetType, state);
+            createAssetDropdown(stateAssets.get(state), pathDictionary[state]);
+        select.val(pathDictionary[state]);
```
juliexxia: maybe make constant?

docs/import/manage_disaster.js
```diff
@@ -675,41 +683,87 @@ function toggleState(known) {
   }
 }
 
+const scoreAssetTypes = [
+  ['poverty', ['snap_data', 'paths'], 'Poverty'],
```
juliexxia: maybe do some magic to combine the 0 and 2 index here?

docs/import/manage_disaster.html
janakdr: Maybe better? Also fixed missing div, I think

docs/import/manage_disaster.js
janakdr: Nope, good point.

docs/import/manage_disaster.js
janakdr: select.val shouldn't still be there, removed.

docs/import/manage_disaster.js
janakdr: What do you suggest? Putting them in a sub-array? Or putting them closer together in this array? I'd rather not make a big structured thing like {key: ..., path: [...], display_name: ...}

juliexxia: Oh,  I meant since [2] seems to be [0] with a capitalized first letter maybe we can just store one and derive the other from it. Unless you think that pattern will change going forward?

docs/import/list_ee_assets.js
```diff
@@ -45,7 +45,8 @@ function getStatesAssetsFromEe(states) {
  * of asset path to type for the given disaster.
  */
 function getDisasterAssetsFromEe(disaster) {
-  return listEeAssets(eeLegacyPathPrefix + disaster);
+  return listEeAssets(eeLegacyPathPrefix + disaster)
```
juliexxia: We're just returning the asset paths now? Does that mean disasterAssets isn't storing type information anymore? Does this break createDisasterAssetPickers?

docs/import/manage_disaster.js
janakdr: Yeah, I think it's not to be relied on.

docs/import/list_ee_assets.js
janakdr: Look at latest revision :)

docs/import/manage_disaster.html
```diff
@@ -104,12 +104,12 @@
...
         <table id="asset-selection-table">
           <thead>
-          <tr id="score-asset-header-row">
-          </tr>
+          <tr id="score-asset-header-row"><td>Score Assets</td></tr>
           </thead>
           <tbody id="asset-selection-table-body"></tbody>
         </table>
       </div>
+      <div><label for="damage-asset-select">Damage asset:</label><select id="damage-asset-select"></select>
```
juliexxia: nit: add space after 'Damage asset: ' (i think in reality we want padding around all our labels, but understandable if you don't want to futz with that right now)

docs/import/manage_disaster.js
```diff
@@ -557,7 +568,10 @@ function onSetDisaster() {
  */
 function maybeFetchDisasterAssets(disaster) {
   if (!disasterAssets.has(disaster)) {
-    disasterAssets.set(disaster, getDisasterAssetsFromEe(disaster));
+    disasterAssets.set(
+        disaster,
+        getDisasterAssetsFromEe(disaster).then(
+            (result) => Array.from(result.keys())));
```
juliexxia: I think we need to update the doc for disasterAssets?

docs/import/manage_disaster.js
```diff
@@ -675,41 +686,85 @@ function toggleState(known) {
...
  */
 function initializeScoreSelectors(states) {
   const headerRow = $('#score-asset-header-row');
-  const tableBody = $('#asset-selection-table-body');
-  tableBody.empty();
-  headerRow.empty();
 
   // Initialize headers.
-  headerRow.append(createTd().html('Score Assets'));
+  removeAllButFirstFromRow(headerRow);
```
juliexxia: I don't think this line is working, not sure why.

janakdr: Sorry, fixed in the sequel but didn't back port.

PR #257, opened 2019-12-03
Prototype: add ui for non-ee layers

ruthtalbot:
Will clean up if the overall approach is right.

### DIFF
--- a/docs/import/add_layer.js
+++ b/docs/import/add_layer.js
@@ -74,13 +74,17 @@ function prependToTable(layer) {
 
 /**
  * Adds a new non-ee layer to the table and firestore.
- * @param {string} name The display name for the layer
  * @param {number} type The layer type (kml or map tile)
  * @param  {Array<string>} The urls for the layer display
  * @return {Promise<void>} Finishes when the layer information has been
  * written to firestore.
  */
-function processNonEeLayer(name, type, urls) {
-  const layer = {'display-name' : name, 'asset-type': type, 'urls' : urls, 'display-on-load': false};
+function processNonEeLayer(type, urls) {
+  const layer = {
+    'display-name': '',
+    'asset-type': type,
+    'urls': urls,
+    'display-on-load': false
+  };
   return prependToTable(layer);
 }
--- a/docs/import/add_layer.js
+++ b/docs/import/add_layer.js
@@ -3,7 +3,7 @@ import {convertEeObjectToPromise} from '../map_util.js';
 import {createLayerRow} from './manage_layers.js';
 import {getCurrentLayers, updateLayersInFirestore} from './manage_layers_lib.js';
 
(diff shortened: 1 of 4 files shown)


docs/import/manage_layers.html
```diff
@@ -73,6 +73,17 @@
   </div>
   <p>Upload additional assets via the <a href="https://code.earthengine.google.com/" target="_blank"> code editor </a></p>
   <br>
+  <div>
+    <div>Add Other Layer Type</div>
+    <select id="non-eelayer-type">
+      <option value="4">Map Tiles</option>
+      <option value="5">KML</option>
+    </select>
+    <label>Layer Name</label><input id="non-eelayer-name"></input>
```
juliexxia: When we add ee layers we initialize the layer with an empty name. Might be nice to remove this input so it's parallel to what happens with ee assets (add layer -> add name in Layer Name cell)

docs/import/manage_layers.js
```diff
@@ -402,3 +404,17 @@ function createOptionFrom(innerTextAndValue) {
...
 }
+
+/**
+ * Adds a non-ee layer to the map. 
+ */
+function addNonEELayer() {
+  const type = parseInt($('#non-eelayer-type').val());
+  const urls = $('#non-eelayer-urls').val().split('\n');
+
+  processNonEeLayer($('#non-eelayer-name').val(), type, urls);
```
juliexxia: Might be nice to return this promise so we know when it's added (and for easier testing)

docs/import/manage_layers.html
ruthtalbot: Sounds good. Updated!

docs/import/manage_layers.js
ruthtalbot: done

docs/import/add_layer.js
```diff
@@ -74,13 +74,17 @@ function prependToTable(layer) {
 
 /**
  * Adds a new non-ee layer to the table and firestore.
- * @param {string} name The display name for the layer
  * @param {number} type The layer type (kml or map tile)
```
juliexxia: nit: Maybe reference LayerType enum here

docs/import/manage_layers.html
```diff
@@ -33,57 +33,65 @@
...
+          <option value="4">Map Tiles</option>
+          <option value="5">KML</option>
+        </select>
+      </div>
+      <br>
+      <div>
+        <label>Layer Urls</label>
+        <textarea id="non-eelayer-urls" style="width:500px" rows="3"></textarea>
+      </div>
+      <button id="add-non-eelayer">Add Layer</button>
```
juliexxia: Do you want this inside the above div too? or no. I have no opinion either way.

PR #258, opened 2019-12-03
write disaster folder on new disaster write

juliexxia:
Create EE folders for both the disaster and the relevant states.  Wait to automatically toggle to new disaster (or let new disaster appear in picker) until these things have finished.

### DIFF
--- a/docs/import/list_ee_assets.js
+++ b/docs/import/list_ee_assets.js
@@ -1,4 +1,4 @@
-import {eeLegacyPathPrefix, eeStatePrefixLength, legacyStateDir} from '../ee_paths.js';
+import {eeLegacyPathPrefix, legacyStateDir} from '../ee_paths.js';
 import {LayerType} from '../firebase_layers.js';
 
 export {getDisasterAssetsFromEe, getStatesAssetsFromEe};
--- a/docs/import/list_ee_assets.js
+++ b/docs/import/list_ee_assets.js
@@ -13,7 +13,7 @@ function getStatesAssetsFromEe(states) {
   const promises = [];
   for (const state of states) {
     const dir = legacyStateDir + '/' + state;
-      promises.push(listEeAssets(dir).then((result) => [state, result]));
+    promises.push(listEeAssets(dir).then((result) => [state, result]));
   }
   return Promise.all(promises);
 }
--- a/docs/import/list_ee_assets.js
+++ b/docs/import/list_ee_assets.js
@@ -10,31 +10,12 @@ export {getDisasterAssetsFromEe, getStatesAssetsFromEe};
  *     retrieved assets in the form [['WA', {'asset/path': LayerType,...}], ...]
  */
 function getStatesAssetsFromEe(states) {
-  return ee.data.listAssets(legacyStateDir, {}, () => {}).then((result) => {
-    const folders = new Set();
-    for (const folder of result.assets) {
-      folders.add(folder.id.substring(eeStatePrefixLength));
-    }
-    const promises = [];
-    for (const state of states) {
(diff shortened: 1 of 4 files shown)


docs/import/manage_disaster.js
```diff
@@ -635,34 +637,67 @@ function writeNewDisaster(disasterId, states) {
...
+ * TODO: add status bar for when this is finished.
+ *
+ * @param {string} dir asset path of folder to create
+ * @return {Promise<void>} resolves when after the directory is created and
+ * set to world readable.
+ */
+function getCreateFolderPromise(dir) {
+  return new Promise((resolve) => {
+    ee.data.createFolder(dir, false, () => {
+      ee.data.setAssetAcl(dir, {all_users_can_read: true});
```
janakdr: I think maybe we should make this async as well? Then can just pass resolve as the callback, and one-line everything :)

docs/import/manage_disaster.js
```diff
@@ -635,34 +637,67 @@ function writeNewDisaster(disasterId, states) {
...
+      }
+    });
+    if (!added) disasterPicker.append(createOptionFrom(disasterId));
+    toggleState(true);
 
-  disasterPicker.val(disasterId).trigger('change');
+    disasterPicker.val(disasterId).trigger('change');
+  });
 
   return disasterCollectionReference()
```
janakdr: maybe should return the result of all of the promises? So above do "const eePromiseResult = Promise.all(folderCreationPromises).then(()..." and here do "return Promise.all(eePromiseResult, disasterCollectionReference().doc(disasterId).set(currentData)).then(() => true)?

docs/import/list_ee_assets.js
```diff
@@ -10,31 +10,12 @@ export {getDisasterAssetsFromEe, getStatesAssetsFromEe};
  *     retrieved assets in the form [['WA', {'asset/path': LayerType,...}], ...]
  */
 function getStatesAssetsFromEe(states) {
```
janakdr: As a follow-up, we could fetch all disasters' states' data eagerly, instead of doing it on demand.

juliexxia: Added a TODO

docs/import/manage_disaster.js
```diff
@@ -628,39 +631,75 @@ function addDisaster() {
...
-  if (!added) disasterPicker.append(createOptionFrom(disasterId));
-  toggleState(true);
 
-  disasterPicker.val(disasterId).trigger('change');
+  return Promise.all(
+      eePromisesResult,
+      disasterCollectionReference()
+          .doc(disasterId)
+          .set(currentData)
+          .then(() => true));
```
janakdr: subtle point, but you need to but the .then(() => true) after the Promise.all() closing parens, not inside, so it actually returns a boolean Promise as per the contract.

docs/import/manage_disaster.js
```diff
@@ -628,39 +631,75 @@ function addDisaster() {
...
+ * @param {string} dir asset path of folder to create
+ * @return {Promise<void>} resolves when after the directory is created and
+ * set to world readable.
+ */
+function getCreateFolderPromise(dir) {
+  return new Promise(
+      (resolve) => ee.data.createFolder(
+          dir, false,
+          () => ee.data.setAssetAcl(
+              dir, {all_users_can_read: true}, () => resolve())));
```
janakdr: nit: I think you can just put resolve here, instead of () => resolve(), since it's already being passed around as a lambda

docs/import/manage_disaster.js
juliexxia: Ah, I had actually changed the doc but this is better. Done.

PR #259, opened 2019-12-03
Add remaining score selectors, write to Firestore

janakdr:
Data is saved to Firestore using the same code as for manage_layers.js. Remaining score assets (TIGER shapefiles and MS buildings) can be modified.

Remaining work is now just to disable "kick off" button until all assets specified.


janakdr:
Reminder to myself to change upstream branch to master once merged.

docs/import/manage_layers_lib.js
```diff
@@ -79,51 +79,16 @@ function onUpdate(event, property, fxn) {
...
-          case STATE.QUEUED_WRITE:
-            loadingElementFinished(writeWaiterId);
-            return updateLayersInFirestore();
-          case STATE.SAVED:
-            console.error('Unexpected layer write state');
-            return null;
-        }
-      });
+  return updateDataInFirestore(
+      () => ({layers: getCurrentLayers()}),
```
juliexxia: do you need parens around curlies?

docs/import/update_firestore_disaster.js
```diff
@@ -0,0 +1,57 @@
...
+ * Writes the current state of a disaster's data to firestore.
+ * @param {Function} dataSupplier Function that returns data to be written for
+ *     current disaster
+ * @param {Function} startCallback Function that should be called before a write
+ *     starts (to indicate loading progress, for instance)
+ * @param {Function} finishCallback Function called whenever a write ends
+ * @return {?Promise<void>} Returns when finished writing or null if it just
+ * queued a write and doesn't know when that will finish.
+ */
+function updateDataInFirestore(dataSupplier, startCallback, finishCallback) {
```
juliexxia: maybe make startCallback and finishCallback optional?

docs/import/manage_disaster.js
```diff
@@ -804,9 +829,25 @@ function createAssetDropdown(
...
 }
 
+/**
+ * Handles the user selecting an asset for one of the possible score types.
+ * @param {Event} event Event, used to identify the target
+ * @param {Array<string>} propertyPath path to property inside asset data. We
+ *     set this value by setting the parent's attribute to the target's value
+ */
+function handleScoreAssetSelection(event, propertyPath) {
+  const parentProperty = getElementFromPath(propertyPath.slice(0, -1));
```
juliexxia: Maybe add some logic commentary?

Also, wondering how intuitive it is for changing the value of the drop down to trigger a write. I guess changing the disaster drop down toggles something but that's visually clear where for other drop downs you have to click a button to trigger anything. Maybe a TODO for later to have some visual indication that this changed something.

docs/import/manage_disaster.js
```diff
@@ -550,13 +552,23 @@ function onSetDisaster() {
...
       if (getDisaster() === currentDisaster &&
           !processedCurrentDisasterSelfAssets) {
         // Don't do anything unless this is still the right disaster.
         initializeDamageSelector(assets);
         processedCurrentDisasterSelfAssets = true;
       }
+    };
+    // Handle errors in disaster asset retrieval by just emptying out.
+    disasterAssets.get(currentDisaster).then(disasterLambda, (err) => {
+      if (err &&
```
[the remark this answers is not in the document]
janakdr: I think you might mean pre-damage :) This is what happens without your PR, since the disaster folder doesn't exist yet. We can delete this with your PR, or keep it if we think they might open this disaster in a new page when the EE folder creation hasn't succeeded yet. Since it doesn't actually cost any processing time, I'm ok with leaving it in.

docs/import/manage_disaster.js
janakdr: This is by analogy with the manage_layers page, where changing the value of drop-downs also triggers writes, no?

Added logic explanation.

docs/import/manage_layers_lib.js
janakdr: Yes, otherwise Javascript interprets the curlies as denoting the body of the function (since this is a lambda) and gets very confused.

docs/import/update_firestore_disaster.js
janakdr: My suspicion is that they won't be, because we'll need to know when all writes have finished for deletions. So I think not worth it.

PR #260, opened 2019-12-04
Offer option of setting map bounds manually when damage asset not supplied

janakdr:
Also patch little bug where we were setting value in Firestore to "None" if select had "None" chosen as an option. I kicked off asset creation with this on a little section of Panama City beach. You can see the results on our map if you change the disaster to 1999-janak.

Finally added a bit of test coverage.


docs/import/manage_disaster.html
```diff
@@ -110,6 +110,8 @@
         </table>
       </div>
       <div><label for="damage-asset-select">Damage asset: </label><select id="damage-asset-select"></select>
+        <div id="map-bounds-div" hidden>
+         OR (before damage asset available)<br/><label for="map-bounds-sw">Southwest damage bounds, in "lat, lng" format: </label><input id="map-bounds-sw"><br/><label for="map-bounds-sw">Northeast damage bounds, in "lat, lng" format: </label><input id="map-bounds-ne"></div>
```
juliexxia: I think it would be helpful to have a literal example the way we do (probably unnecessarily) for the add disaster fields.

docs/import/manage_disaster.js
```diff
@@ -829,25 +853,37 @@ function createAssetDropdown(
...
 /**
- * Handles the user selecting an asset for one of the possible score types.
- * @param {Event} event Event, used to identify the target
+ * Adds the default change handler, which updates our internal data (and
+ * Firestore) when this element changes.
+ * @param {JQuery<HTMLElement>} elt
+ * @param {Array<string>} propertyPath The path to the value of this element
+ * @return {JQuery<HTMLElement>} The passed-in element, for chaining
+ */
+function addChangeHandler(elt, propertyPath) {
```
juliexxia: nit: can we make this function name more specific like addScoreAssetChangeHandler

docs/import/manage_disaster.js
```diff
@@ -770,8 +778,24 @@ function initializeScoreSelectors(states) {
...
+    const val = $(event.target).val();
+    hasValue(val) ? mapBoundsDiv.hide() : mapBoundsDiv.show();
+    handleScoreAssetSelection(val, damagePropertyPath);
+  });
+  const swPath = ['map_bounds_sw'];
+  const nePath = ['map_bounds_ne'];
+  const swInput = $('#map-bounds-sw');
+  swInput.val(getElementFromPath(swPath));
+  addChangeHandler(swInput, swPath);
+  const neInput = $('#map-bounds-ne');
```
juliexxia: Is it weird that if you input one and not the other coord it gets stored?

docs/import/manage_disaster.js
```diff
@@ -829,25 +853,37 @@ function createAssetDropdown(
...
+ */
+function addChangeHandler(elt, propertyPath) {
+  return elt.on(
+      'change',
+      (event) =>
+          handleScoreAssetSelection($(event.target).val(), propertyPath));
+}
+/**
+ * Handles the user entering a value into score-related input
+ * @param {string} val Value of input. 'None'/empty are treated like null (ugh)
```
juliexxia: TODO to not have this? I think we can just have empty without None probably?

docs/import/manage_disaster.html
janakdr: This look ok?

docs/import/manage_disaster.js
janakdr: Don't think so? You have to find them one at a time (I'm hoping someone can write documentation about how to do this if we don't have it already (https://support.google.com/maps/answer/18539?co=GENIE.Platform%3DDesktop&hl=en) and link it here? @ruthtalbot ), so there's going to be a delay in inputting them. If only one is input, we won't allow processing.

docs/import/manage_disaster.js
janakdr: the map bounds aren't "assets" per se, but are handled here. Changed name here and below to use "asset data", which is still kind of a lie, I guess.

docs/import/manage_disaster.js
janakdr: You're right, got rid of it.

docs/import/manage_disaster.js
```diff
@@ -864,26 +866,24 @@ function createAssetDropdown(
...
   return elt.on(
       'change',
-      (event) =>
-          handleScoreAssetSelection($(event.target).val(), propertyPath));
+      (event) => handleAssetDataChange($(event.target).val(), propertyPath));
 }
 /**
  * Handles the user entering a value into score-related input
- * @param {string} val Value of input. 'None'/empty are treated like null (ugh)
+ * @param {string} val Value of input. empty strings are treated like null (ugh)
```
juliexxia: Is this doc still right?

janakdr: i think so? if val evaluates to false we set it to null. Fixed a bit more, though, because we don't want false to turn into null.

PR #261, opened 2019-12-04
Disable kick-off button until all assets specified

janakdr:
Make buttons gray when disabled, and make this button not capitalize, since it has a lot of text in it. Hide the button until all asset retrieval is done.

There's a subtle issue here if Firestore's value for the asset for some score factor is not present in EE. Currently, we won't display that value in the select, but we don't overwrite the Firestore value or our local value until the user actually changes the dropdown, so we could incorrectly think the user had specified all assets, even though one doesn't exist. Not sure if worth addressing, since should be pretty hard for user to do that (easiest way would be to delete an asset, reload page, then not notice a select wasn't selected anymore). Not new in this PR either, although potentially more confusing because page will say we're good to go. Since we'll always be vulnerable to user deleting assets and *not* reloading page (in which case we'll show stale data forever), I think we can punt.


cypress/integration/unit_tests/manage_disaster_test.js
```diff
@@ -352,6 +443,66 @@ describe('Unit tests for manage_disaster.js', () => {
...
+   * with damage/map-bounds-related HTML elements.
+   * @return {Cypress.Chainable<Document>}
+   */
+  function setAssetDataAndCreateDamageInputsForScoreValidationTests() {
+    const currentData = createDisasterData(['NY']);
+    currentData['asset_data']['map_bounds_sw'] = '0, 1';
+    disasterData.set(getDisaster(), currentData);
+    return cy.document().then((doc) => {
+      // Lightly fake out jQuery so that we can use Cypress selectors. Might not
+      // work if manage_disaster.js starts doing fancier jQuery operations.
```
juliexxia: Why not just use jquery selectors here and below?

cypress/integration/unit_tests/manage_disaster_test.js
```diff
@@ -352,6 +443,66 @@ describe('Unit tests for manage_disaster.js', () => {
...
+  function setUpAssetValidationTests() {
+    // Clear out any modifications we've done to the document.
+    cy.visit('test_utils/empty.html');
+    return setAssetDataAndCreateDamageInputsForScoreValidationTests().then(
+        (doc) => {
+          cy.stub(document, 'createElement')
+              .callsFake((tag) => doc.createElement(tag));
+          const tbody = doc.createElement('tbody');
+          tbody.id = 'asset-selection-table-body';
+          doc.body.appendChild(tbody);
```
juliexxia: createAndAppend a potentially useful operation (i think in import_test_utils)

cypress/integration/unit_tests/manage_disaster_test.js
```diff
@@ -352,6 +443,66 @@ describe('Unit tests for manage_disaster.js', () => {
...
   });
+
+  /**
+   * Sets up fresh asset data (no feature collections), and prepares fake page
+   * with damage/map-bounds-related HTML elements.
+   * @return {Cypress.Chainable<Document>}
+   */
+  function setAssetDataAndCreateDamageInputsForScoreValidationTests() {
+    const currentData = createDisasterData(['NY']);
+    currentData['asset_data']['map_bounds_sw'] = '0, 1';
```
juliexxia: just setting sw? Asking because above you say "We haven't set anything"

juliexxia:
Approving since all my comments are test-related and not that major!

cypress/integration/unit_tests/manage_disaster_test.js
```diff
@@ -217,6 +191,123 @@ describe('Unit tests for manage_disaster.js', () => {
...
+    // Specifying assets for income, both states, one type, gets rid of income
+    // from message.
+    setFirstSelectInScoreRow(1);
+    getFirstTdInScoreRow(1).next().next().find('select').select('wy1').blur();
+    cy.get('#process-button')
+        .should(
+            'have.text',
+            'Missing asset(s): Poverty [WY], SVI [NY, WY], Census TIGER ' +
+                'Shapefiles [NY, WY], Microsoft Building Shapefiles [NY, WY],' +
+                'and must specify either damage asset or map bounds');
```
juliexxia: missing a space between these last two lines

cypress/integration/unit_tests/manage_disaster_test.js
janakdr: we need to append to doc.body, so would need more mocking. Good idea for future.

cypress/integration/unit_tests/manage_disaster_test.js
janakdr: Make buttons gray when disabled, and make this button not capitalize, since it has a lot of text in it. Hide the button until all asset retrieval is done.

There's a subtle issue here if Firestore's value for the asset for some score factor is not present in EE. Currently, we won't display that value in the select, but we don't overwrite the Firestore value or our local value until the user actually changes the dropdown, so we could incorrectly think the user had specified all assets, even though one doesn't exist. Not sure if worth addressing, since should be pretty hard for user to do that (easiest way would be to delete an asset, reload page, then not notice a select wasn't selected anymore). Not new in this PR either, although potentially more confusing because page will say we're good to go. Since we'll always be vulnerable to user deleting assets and not reloading page (in which case we'll show stale data forever), I think we can punt.

cypress/integration/unit_tests/manage_disaster_test.js
janakdr: rephrased above.

PR #266, opened 2019-12-04
Disable michael in tests

janakdr:
With this, we can re-enable github status checks.


janakdr:
Had to disable polygon_draw_test#calculateDamage

PR #268, opened 2019-12-04
Get rid of default "grab/pointer" styling in manage_disaster table

janakdr:


docs/import/manage_disaster.html
```diff
@@ -27,6 +27,11 @@
   <!-- Load our styles. -->
   <link rel="stylesheet" href="../style.css">
   <link rel="stylesheet" href="manage.css">
+  <style>
```
juliexxia: can we stick this in a .css file somewhere?

janakdr: Went all in.

PR #274, opened 2019-12-04
Strip feature collections before rendering

juliexxia:


juliexxia:
This actually doesn't seem worth it/possible. Testing with the harvey damage dataset, we run into earth engine memory limit errors when trying to evaluate new reduced-to-one-property feature collection. Probably something to do with earth engine optimizing for assets that are already generated and stored in ee vs newly created ones.

PR #285, opened 2019-12-05
Pull actual asset-creation functions out of manage_disaster.js

janakdr:
File was getting unwieldy. Only change (that I can remember) is making createScoreAsset (formerly run) return null instead of false if damageEnvelope was null, since that's what createScoreAsset returns when there's an error.


docs/import/manage_disaster.js
```diff
@@ -1,12 +1,10 @@
 import {eeLegacyPathPrefix, legacyStateDir} from '../ee_paths.js';
 import {LayerType} from '../firebase_layers.js';
 import {disasterCollectionReference} from '../firestore_document.js';
-import {blockGroupTag, buildingCountTag, damageTag, geoidTag, incomeTag, snapPercentageTag, snapPopTag, sviTag, totalPopTag, tractTag} from '../property_names.js';
-import {getDisaster, getScoreAsset} from '../resources.js';
+import {getDisaster} from '../resources.js';
```
juliexxia: super nit: appears to not be new but get rid of this /n?

PR #289, opened 2019-12-06
Make kml layers unclickable, so user can interact with the map below

ruthtalbot:
This does require removing those info windows, but Alex didn't seem to find them interesting/helpful anyway. 

We could do something more complicated where we add an event listener and determine if there's something clickable below, but not sure it's worth it given he didn't seem to care about the kml info popup.


janakdr:
Closes #281

PR #290, opened 2019-12-07
Move most of click_feature_test to a unit test and make some production changes along the way

janakdr:
Production changes:
- ChartWrapper, which we were using instead of a bare google.visualization.Table, doesn't make the Table available until after it's ready. While that is generally instant in production, it takes ~7 seconds in a unit test, maybe because of library initialization latency that's hidden in prod by other things(?). However, using the bare Table works just as well for our purposes, and is actually a bit more compact. I think this is also a potential source of click_feature_test's flakiness: on the rare occasions when the ChartWrapper wasn't immediately ready, clicking features wouldn't work, because we'd have a null table. I did try the recommended mechanism of adding a 'ready' listener to ChartWrapper, but that seemed to add a lot of unnecessary latency: I could see the table on the page, but the ready handler wasn't triggered for several seconds afterwards, at least in tests.
- Instead of passing the Table and list of features around for consumers who want to select it, provide a function that encapsulates it. That reduces code duplication outside as well, and keeps knowledge of the table-drawing internals inside draw_table.js.
- Have that function be returned (in a Promise) by draw_table, since it wasn't returning anything before.
- Get the column names for features directly from tableHeadings, instead of indirectly from the list
- Deleted highlight_feature.js, which has been dead for months (as opposed to highlight_features.js)

Test changes:
- Newly added unit test supersedes previous version: it exercises all previous codepaths but does strictly more. It also replaces most of integration test, since it has a real table and real map to work with. It can even assert more: that there really is a polygon being rendered on the data layer of the map.
- Test map creation is now a bit more realistic, in that it has a non-zero-sized div, and a reasonable zoom (it was the whole world before, basically, repeated twice, which explains why checkbox_test was requesting tiles twice). You can see the map in Cypress during tests, which is nice.
- Integration click_feature_test is left around because I didn't want to deal with update functionality as well in this unit test. Also probably good to have one integration test case.
- Added support for google.charts to script_loader.


docs/run.js
```diff
@@ -72,26 +72,32 @@ function createAndDisplayJoinedData(
...
+
+/**
+ * Invokes {@link drawTable} with the appropriate callbacks to set up click
+ * handlers for the map.
+ * @param {Promise<Array<GeoJson.Feature>>} processedData
+ * @param {google.maps.Map} map
+ * @param {string} scoreAsset EE path to score asset FeatureCollection
+ */
+function drawTableAndSetUpHandlers(processedData, map, scoreAsset) {
+  drawTable(processedData, (features) => highlightFeatures(features, map, true))
```
juliexxia: maybe define the callback inside drawTable since drawTable is only being called once? Makes easier to understand code inside drawTable.

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -1,90 +1,201 @@
...
+ * @param {number} north North-most (largest latitude) coordinate of feature
+ * @return {ee.Feature}
+ */
+function createFeatureFromCorners(west, south, east, north) {
+  const polygonCoordinates =
+      [west, south, west, north, east, north, east, south];
+  let result = ee.Feature(ee.Geometry.Polygon(polygonCoordinates));
+  for (let i = 0; i < tableHeadings.length; i++) {
+    result = result.set(
+        tableHeadings[i], polygonCoordinates[i % polygonCoordinates.length]);
```
juliexxia: Setting each column in the table to some function of the polygon coords?

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -1,90 +1,201 @@
...
+  }
 
-    clickFeature(0, 0, null, 'mockAsset', tableApi, tableData);
-    clickFeature(0, 0, null, 'mockAsset', tableApi, tableData);
+  /**
+   * Gets all features currently displayed on map.
+   * @return {Cypress.Chainable<Array<google.maps.Data.Feature>>}
+   */
+  function getDataFeatures() {
+    return cy.wrap(null).then(() => {
```
juliexxia: what does the cy.wrap(null) do again? mind leaving a comment?

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -1,90 +1,201 @@
...
+   */
+  function setUpPage() {
+    const loadingFinishedPromise =
+        new Promise((resolve) => loading.loadingElementFinished = (id) => {
+          if (id === tableContainerId) resolve();
+        });
+    createGoogleMap().then((mapResult) => map = mapResult);
+    cy.document().then((doc) => {
+      // Lightly fake out prod document access.
+      cy.stub(document, 'getElementById')
```
juliexxia: we should probably standardize this eventually. We've been doing different things. maybe add a todo?

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -1,90 +1,201 @@
...
+      doc.body.appendChild(containerDiv);
+      const tableDiv = doc.createElement('div');
+      tableDiv.id = 'table';
+      containerDiv.appendChild(tableDiv);
+      drawTableAndSetUpHandlers(
+          convertEeObjectToPromise(scoredFeatures).then((fc) => fc.features),
+          map, features);
+    });
+    cy.wrap(loadingFinishedPromise);
+    return assertNoSelection();
```
juliexxia: why return if not being chained?

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -1,90 +1,201 @@
...
+  }
 
-    mockTable.verify();
+  it('clicks on a feature on the map, then unclicks it', () => {
+    setUpPage();
+    clickAndVerifyBlockGroup();
+    cy.wait(waitBeforeClick);
+    cy.get('#test-map-div').click(500, 100);
```
juliexxia: Sort of tough that we're using long lat for creation and pixels for clicking. Assuming you just experimented until you found good numbers?

docs/run.js
janakdr: Great idea, done.

cypress/integration/unit_tests/click_feature_test.js
janakdr: someone might chain off of it. It's also a very good signal that this function does things in the "Cypress phase". The convention I'm using (described more explicitly in my follow-up PR) is that a function that does things in the Cypress phase returns a Chainable, while a function that does things in normal execution order returns a normal object, since why else would you be calling it.

cypress/integration/unit_tests/click_feature_test.js
janakdr: Yup, same as with integration tests.

cypress/integration/unit_tests/click_feature_test.js
janakdr: This is a fundamental enough concept that I think this isn't the best place for it. Maybe a README somewhere, though I don't think we have it yet. (I do this a lot more in the follow-up.) Any suggestions for a central place? Or I could create a library function "executeDuringCypressPhase()" that takes a callback, and document there. How about that?

cy.wrap(null).then() makes the code inside the then() only execute in the "Cypress phase" of execution, not during the initial phase. Without the .then(), we'd try to access the map variable even before it was created (because the createMap() function doesn't actually create the map until the Cypress phase).

cypress/integration/unit_tests/click_feature_test.js
janakdr: Yeah, just an arbitrary value that will vary based on the passed-in polygon.

cypress/integration/unit_tests/click_feature_test.js
janakdr: Added a function in commands.js. Will migrate downstream PRs to use it.

cypress/support/commands.js
```diff
@@ -24,3 +26,43 @@ Cypress.Commands.add('awaitLoad', (divIds) => {
...
+ * an internal Cypress "queue". After all lines have been executed, Cypress
+ * knows the full queue. At that point, the "Cypress phase" starts. Cypress
+ * actually executes each command in the queue. After each command completes, it
+ * will execute any `.then` blocks that were chained onto it. Thus, in the
+ * following program:
+ * ```
+ * console.log('message 1');
+ * cy.wrap(console.log('message 2')).then(() => console.log('message 4'));
+ * console.log('message 3');
+ * cy.get('#id').then(() => console.log('message 5'));
```
juliexxia: can you add a cy.wrap(console.log(...)) between the third and fourth lines?

cypress/support/commands.js
```diff
@@ -24,3 +26,43 @@ Cypress.Commands.add('awaitLoad', (divIds) => {
...
+ * Cypress.Chainable} can be assumed to execute its instructions during the
+ * Cypress phase. Only functions that do not return a {@link Cypress.Chainable}
+ * execute during the initial phase.
+ *
+ * For more information, see
+ * https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Asynchronous
+ * @param {Function} lambda Code to be executed during Cypress phase.
+ * @return {Cypress.Chainable<any>}
+ */
+function cyQueue(lambda) {
```
juliexxia: So would it be accurate to say it only makes sense to call this function as the first thing in a method? (test or utility?)

docs/draw_table.js
```diff
@@ -87,8 +86,8 @@ function renderTable(list, features, selectTableCallback, selectorReceiver) {
 
   google.visualization.events.addListener(
       table, 'select',
-      () => selectTableCallback(
-          table.getSelection().map((elt) => features[elt.row])));
+      (features) => highlightFeatures(
```
juliexxia: is features passed to highlight features no longer the features param that's being passed in to renderTable?

cypress/integration/unit_tests/click_feature_test.js
janakdr: geoid needs to be unique, so used "west" coordinate. After that, seemed just as easy as not to initialize with remaining coordinates. Added a comment.

cypress/support/commands.js
janakdr: Sure, and added even more.

cypress/support/commands.js
janakdr: Good point, added.

docs/draw_table.js
janakdr: Great catch! integration_tests/highlight_features_test failed with this bug, so I guess we have some coverage :)

docs/click_feature.js
```diff
@@ -20,6 +20,7 @@ function clickFeature(lng, lat, map, featuresAsset, tableSelector) {
   const blockGroups = ee.FeatureCollection(featuresAsset).filterBounds(point);
   const selected = blockGroups.first();
   selected.evaluate((feature, failure) => {
+    console.log(feature);
```
juliexxia: debugging? here and below?

Issue #282, opened 2019-12-05
Add links to documentation


RamV13:
This was started in #267 but we can make this better with links to specific sections and maybe even embedding the doc content.

janakdr:
Should we leave open for embedding doc content possibility? I think @ruthtalbot has figured out how to put "preview" mode in an iframe?

RamV13:
@ruthtalbot please share when you get a chance and I'll take this on

janakdr:
I'm taking a look now. Ruth says that if you publish to the web as an embedded doc, it works. So for example https://docs.google.com/document/d/e/2PACX-1vQj4adOV3Dsw9sBiBRcFlqmvAi3blvE3EC8hvv2CO9iH98URxfPGkZ6B6OPWm29PSZQTUqMq32hpuY4/pub?embedded=true#h.bvkreu5dkxxn links to the polygon section. Give me a few more minutes :)

janakdr:
Mostly failed result of experimenting: https://github.com/givedirectly/Google-Partnership/pull/373. Ram, I'm sure you can do better than me, but not sure if Google docs formatting will ever look acceptable in a pop-up :(

RamV13:
Yup I see what you mean this doesn't look ideal but I'm not sure what else can be done with Google docs alone.

janakdr:
For reference, below is what my terrible attempt looks like. @ruthtalbot do you think it's worth pursuing this, or not worth trying to embed google docs?

![image](https://user-images.githubusercontent.com/10134896/71607478-e4d41c80-2b47-11ea-9cef-3a0db90cf926.png)

ruthtalbot:
Sorry, don't know why I didn't see this Monday -- agree that if it's between that and just links, I'd prefer links. Ram if you can get it more usable without sinking much time into it, embedded still feels useful but if not lets keep links. 

I can ask someone on Docs if there's any easy way of doing this.

PR #291, opened 2019-12-07
Adding help links

RamV13:


docs/create_map.js
```diff
@@ -35,6 +36,20 @@ function createMap(firebasePromise) {
   });
   setUpPolygonDrawing(map, firebasePromise);
 
+  // Add the help link.
+  const helpContainer = document.createElement('div');
+  helpContainer.style.padding = '6px';
+  const helpLink = document.createElement('a');
+  helpLink.href = POLYGON_HELP_URL;
+  helpLink.target = '_blank';
```
janakdr: Should we set a common help target window?

docs/navbar.js
```diff
@@ -66,3 +70,17 @@ function loadNavbarWithPicker(
 function getUrlUnderDocs(pathFragment) {
   return import.meta.url.replace(/navbar\.js$/, pathFragment);
 }
+
+/**
+ * Gets the url for the help section relevant to the current page
+ * @return {string}
+ */
+function getHelpUrl() {
+  if (window.location.pathname === '/') {
```
janakdr: Are these paths ok when running under a subdirectory?

docs/create_map.js
RamV13: We have it separately hardcoded in the HTML and maybe we'd want to make this context specific so I'm not sure if it would really be worth having a constant for this.

docs/navbar.js
RamV13: Updated the logic to handle this case

RamV13:
I couldn't find any way to link to the doc in "view" mode either and I guess, as you pointed out, it's probably better to leave it in a link through which they can edit the doc as needed.

docs/create_map.js
janakdr: Ah, I was actually thinking of setting target to a named window and then opening all the help links in the named window. But I think I was misremembering something to do with frames, and this isn't actually possible.

RamV13:
Resolved #282

PR #292, opened 2019-12-08
Make polygon_draw unit test talk to real Firestore and move integration tests there

janakdr:
Leave two integration tests, one to test integration with the checkbox, and one to test badly formed polygons. First one also does some basic integration test checking.

Ported existing unit tests to talk to real Firestore and moved integration tests over. There are the usual subtleties around the order of execution in Cypress. I tried to make it as safe as possible by cy-wrapping all possible helper functions. The remaining ones are returning values to assert on, so the assertions are what have to be wrapped.

Added tests for dragging polygons, which we can simulate now because we have access to the actual polygons.

Since polygon_draw_test was the only test that used the Firestore library without actually talking to Firestore, and now it talks to Firestore, made the unit test Firestore initialization unconditional.

Because our tests are now so much faster, un-parallelize them. Because the unit tests can run on Electron and Electron's video recording works properly on Travis and the unit tests are doing enough that video recordings are useful, break tests into unit and integration, and run unit tests with Electron and video recording.

Also sneak in a change to our push hook so lint can be run on its own without committing.

### DIFF
--- a/cypress/integration/unit_tests/polygon_draw_test.js
+++ b/cypress/integration/unit_tests/polygon_draw_test.js
@@ -1,3 +1,4 @@
+import {mapContainerId, writeWaiterId} from '../../../docs/dom_constants.js';
 import * as loading from '../../../docs/loading.js';
 import {initializeAndProcessUserRegions, setUpPolygonDrawing, StoredShapeData, transformGeoPointArrayToLatLng, userShapes} from '../../../docs/polygon_draw.js';
 import {setUserFeatureVisibility} from '../../../docs/popup.js';
@@ -24,17 +25,13 @@ const defaultData = {
   totalHouseholds: 1,
   notes: '',
 };
-const event = new Event('overlaycomplete');
 
 describe('Unit test for ShapeData', () => {
   loadScriptsBeforeForUnitTests('ee', 'maps', 'firebase', 'jquery');
   let damageCollection;
   before(() => {
-    // Stub out loading update attempts: they pollute console with errors.
-    loading.addLoadingElement = () => {};
-    loading.loadingElementFinished = () => {};
     // Wrap StoredShapeData#update so that we can access the Promise it returns.
-    // See also getCurrentUpdatePromiseInCypress.
+    // See also getCurrentUpdatePromiseInCypressAndAssertWrites.
     StoredShapeData.prototype.innerUpdate = StoredShapeData.prototype.update;
     StoredShapeData.prototype.update = function() {
       currentUpdatePromise = this.innerUpdate();
@@ -50,7 +47,19 @@ describe('Unit test for ShapeData', () => {
   // wrapping update call.
   let currentUpdatePromise = null;
 
+  let loadingStartedStub;
+  let loadingFinishedStub;
+  let expectedLoadingFinished = 1;
   beforeEach(() => {
+    // Track writes.
+    const addStub = cy.stub(loading, 'addLoadingElement');
+    loadingStartedStub = addStub.withArgs(writeWaiterId);
+    const finishStub = cy.stub(loading, 'loadingElementFinished');
+    loadingFinishedStub = finishStub.withArgs(writeWaiterId);
+    expectedLoadingFinished = 1;
+    // Stub out map updates, not useful for us.
+    addStub.withArgs(mapContainerId);
+    finishStub.withArgs(mapContainerId);
     // Default polygon intersects feature1 and feature2, not feature3.
     const feature1 = ee.Feature(
         ee.Geometry.Polygon(0, 0, 0, 10, 10, 10, 10, 0),
@@ -71,7 +80,7 @@ describe('Unit test for ShapeData', () => {
     // Use our custom EarthEngine FeatureCollections.
     cy.stub(resourceGetter, 'getScoreAsset').returns(scoreCollection);
 
-    return setUpPage();
+    setUpPage();
   });
 
   /**
@@ -88,6 +97,8 @@ describe('Unit test for ShapeData', () => {
         .then(() => {
           userRegionData.clear();
           return initializeAndProcessUserRegions(map, Promise.resolve({
+            // Normally damage_asset_path is a string, but code tolerates just
+            // putting an ee.FeatureCollection in.
             data: () => ({asset_data: {damage_asset_path: damageCollection}}),
           }));
         });
@@ -98,12 +109,15 @@ describe('Unit test for ShapeData', () => {
     pressPopupButton('close');
   });
 
-  it('Updates notes', () => {
+  it('Updates notes and shape', () => {
     drawPolygonAndClickOnIt();
     pressPopupButton('edit');
-    cy.get('.notes').type(notes);
+    // Clone path and edit.
+    const newPath = JSON.parse(JSON.stringify(path));
+    newPath[0].lng = 0.5;
+    cy.get('.notes').type(notes).then(() => getFirstFeature().setPath(newPath));
     pressButtonAndWaitForPromise('save');
-    assertOnFirestoreAndPopup(path, withNotes(notes));
+    assertOnFirestoreAndPopup(newPath, withNotes(notes));
   });
 
   it('Deletes polygon', () => {
@@ -120,15 +134,16 @@ describe('Unit test for ShapeData', () => {
       expect(confirmStub).to.be.calledOnce;
       confirmStub.resetHistory();
     });
-    assertOnFirestoreAndPopup(path, defaultData);
+    assertOnFirestoreAndPopup(path);
+    // Reload the page: polygon should still be there.
     setUpPage();
-    assertOnFirestoreAndPopup(path, defaultData);
+    assertOnFirestoreAndPopup(path);
     pressPopupButton('delete').then(() => {
       expect(confirmStub).to.be.calledOnce;
       confirmValue = true;
       confirmStub.resetHistory();
     });
-    assertOnFirestoreAndPopup(path, defaultData);
+    assertOnFirestoreAndPopup(path);
     deletePolygon(confirmStub);
   });
 
@@ -141,7 +156,7 @@ describe('Unit test for ShapeData', () => {
     pressPopupButton('delete').then(() => expect(confirmStub).to.be.calledOnce);
     cy.get('#test-map-div').click();
     cy.get('#test-map-div').should('not.contain', 'damage count');
-    getCurrentUpdatePromiseInCypress()
+    getCurrentUpdatePromiseInCypressAndAssertWrites()
         .then(() => userShapes.get())
         .then((querySnapshot) => {
           expect(querySnapshot).to.have.property('size', 0);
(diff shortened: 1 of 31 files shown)


janakdr:
Comment to myself to remember to change upstream branch to master once changes are merged in there.

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -49,142 +67,182 @@ describe('Unit test for ShapeData', () => {
...
+  /**
+   * Initializes the Google Map; sets up the polygon drawing manager; and sets
+   * up user feature handling. No features are expected on first call, since the
+   * test Firestore database is empty.
+   * @return {Cypress.Chainable}
+   */
+  function setUpPage() {
+    return createGoogleMap()
+        .then((mapResult) => map = mapResult)
+        .then(() => setUpPolygonDrawing(map, Promise.resolve(null)))
```
juliexxia: Can this just be Promise.resolve()

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -49,142 +67,182 @@ describe('Unit test for ShapeData', () => {
...
+        .then((drawingManagerResult) => drawingManager = drawingManagerResult)
+        .then(() => {
+          userRegionData.clear();
+          return initializeAndProcessUserRegions(map, Promise.resolve({
+            data: () => ({asset_data: {damage_asset_path: damageCollection}}),
```
juliexxia: this is setting the damage_asset_path to the actual feature collection but in real life it's a string/asset path that resolves to a feature collection. correct?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
+   * matches expected data, and calls {@link assertOnPopup}.
+   * @param {Array<LatLngLiteral>} path
+   * @param {ExpectedData} expectedData
+   * @return {Cypress.Chainable}
    */
-  setCalculatedData(calculatedData) {
-    this.calculatedData = calculatedData;
+  function assertOnFirestoreAndPopup(path, expectedData) {
+    cy.wrap(null)
+        .then(() => expect(StoredShapeData.pendingWriteCount).to.eql(0))
```
juliexxia: can this expect and the subsequent userShapes.get() be in the same then block?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
+      });
+      google.maps.event.trigger(drawingManager, 'overlaycomplete', event);
+    });
+  }
 
   /**
-   * Sets calculatedData property.
-   * @param {Object} calculatedData
+   * Asserts that all Firestore writes have completed, that Firestore data
+   * matches expected data, and calls {@link assertOnPopup}.
```
juliexxia: "matches expected data" sans notes?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
+        .then((querySnapshot) => {
+          expect(querySnapshot).to.have.property('size', 1);
+          const polygonDoc = querySnapshot.docs[0];
+          const firestoreId = polygonDoc.id;
+          expect(transformGeoPointArrayToLatLng(polygonDoc.get('geometry')))
+              .to.eql(path);
+          const calculatedData = Object.assign({}, expectedData);
+          delete calculatedData.notes;
+          expect(polygonDoc.get('calculatedData')).to.eql(calculatedData);
+          expect(polygonDoc.get('notes')).to.eql(expectedData.notes);
```
juliexxia: confused about this copy to calculated data, removal of notes, and then assertion on notes that were copied and deleted

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
-   * @param {Object} calculatedData
+   * Asserts that all Firestore writes have completed, that Firestore data
+   * matches expected data, and calls {@link assertOnPopup}.
+   * @param {Array<LatLngLiteral>} path
+   * @param {ExpectedData} expectedData
+   * @return {Cypress.Chainable}
    */
-  setCalculatedData(calculatedData) {
-    this.calculatedData = calculatedData;
+  function assertOnFirestoreAndPopup(path, expectedData) {
```
juliexxia: could make expectedData optional and default to defaultData

janakdr:
Migrated to use cyQueue instead of cy.wrap(null).then()

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -49,142 +67,182 @@ describe('Unit test for ShapeData', () => {
...
-      expect(ids).to.eql(['my_id']);
-      expect(records).to.eql([{
-        calculatedData: calculatedData,
-        geometry: polygonGeometry,
-        notes: 'new notes',
-      }]);
-      expect(underTest.id).to.eql('my_id');
-      expect(StoredShapeData.pendingWriteCount).to.eql(0);
+  it('Updates notes', () => {
+    drawPolygonAndClickOnIt();
```
juliexxia: maybe also add a polygon shape update here? make it a general update test

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -49,142 +67,182 @@ describe('Unit test for ShapeData', () => {
...
-          calculatedData: calculatedData,
-          geometry: polygonGeometry,
-          notes: 'racing notes',
-        },
-      ]);
-      expect(underTest.id).to.eql('my_id');
-      expect(StoredShapeData.pendingWriteCount).to.eql(0);
+        userShapes.doc = realDocFunction;
+        fakeCalled = true;
+        return realDoc.set(record);
```
juliexxia: Why can't you just stub out readDoc.set (store the function and call it at the end?)

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -1,43 +1,61 @@
...
+ */
+
+const defaultData = {
   damage: 1,
   snapFraction: 0.1,
-  totalHouseholds: 3,
+  totalHouseholds: 1,
+  notes: '',
 };
+const event = new Event('overlaycomplete');
```
juliexxia: I think this is only used in drawPolygon so initialize in there? Or can it be used in the Draws marker test as well?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -49,142 +67,182 @@ describe('Unit test for ShapeData', () => {
...
+
+  it('Draws a polygon, almost closes while editing', () => {
+    const confirmStub = cy.stub(window, 'confirm').returns(false);
+    drawPolygonAndClickOnIt();
+    pressPopupButton('edit');
+    cy.get('.notes').type(notes);
+    pressPopupButton('close').then(() => expect(confirmStub).to.be.calledOnce);
+    pressPopupButton('save');
+    cy.get('#test-map-div').contains(notes).should('be.visible');
+    getCurrentUpdatePromiseInCypress();
```
juliexxia: calling this without doing anything with what it returns will just ensure that the promise is finished before executing more cypress things correct?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
+      expect(firestoreSpy).to.be.calledOnce;
+    });
+    assertOnFirestoreAndPopup(newPath, defaultData);
+  });
+
+  /**
+   * Draws the default polygon and clicks on it. Then returns spies to observe
+   * calculated-data-related calls made to the polygon's popup, and {@link
+   * ee#List} and
+   * {@link userShapes#doc} calls. Because EE is finicky about spying, the
```
juliexxia: nit: formatting got a little wonky?

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
+   * @typedef {sinon.SinonSpy|sinon.SinonStub} Spy
+   * @return {Cypress.Chainable<{popupPendingCalculationSpy: Spy,
+   *     popupCalculatedDataSpy: Spy, eeSpy: Spy, firestoreSpy: Spy}>}
+   */
+  function drawPolygonAndSetUpSpies() {
+    let popupPendingCalculationSpy;
+    let popupCalculatedDataSpy;
+    let firestoreSpy;
+    const dummyObjectForSpyAssertions = {method: () => {}};
+    const eeSpy = cy.spy(dummyObjectForSpyAssertions, 'method');
```
juliexxia: nit: can we call this something more description than method, like forSpying? Idk that's not great either but

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -205,208 +263,466 @@ describe('Unit test for ShapeData', () => {
...
-  const mockPolygon = {};
-  mockPolygon.getMap = () => true;
-  mockPolygon.getPath = () => polyLatLng;
-  return mockPolygon;
-}
+    // Check box again and verify that notes box can now be brought up.
+    setUserFeatureVisibilityInCypressAndAssert(true, true);
+    // Notes not visible yet.
+    cy.get('#test-map-div').contains(notes).should('not.be.visible');
+    cy.wait(500);
```
juliexxia: why do we still need this wait?

cypress/integration/unit_tests/polygon_draw_test.js
janakdr: Discussed offline, and added a comment.

cypress/integration/unit_tests/polygon_draw_test.js
juliexxia: Bump  on this question?

cypress/integration/unit_tests/polygon_draw_test.js
juliexxia: Also fine with a separate test - just didn't think we were testing that anywhere!

cypress/integration/unit_tests/polygon_draw_test.js
```diff
@@ -625,34 +613,43 @@ describe('Unit test for ShapeData', () => {
...
-   * happens, this function will return the value of currentUpdatePromise, which
-   * will be the promise returned from {@link StoredShapeData#update}.
-   * @return {Cypress.Chainable<Promise<void>>}
+   * happens, this function will retrieve the value of currentUpdatePromise,
+   * which will be the promise returned from {@link StoredShapeData#update}.
+   * @return {Cypress.Chainable}
    */
-  function getCurrentUpdatePromiseInCypress() {
-    return cy.wait(0).then(() => currentUpdatePromise);
+  function getCurrentUpdatePromiseInCypressAndAssertWrites() {
```
juliexxia: maybe rename function now that it's not actually returning the promise

cypress/integration/unit_tests/polygon_draw_test.js
janakdr: We were incidentally testing it in the test below that asserted that ee computation happened if geometry changed.

cypress/integration/unit_tests/polygon_draw_test.js
janakdr: Yup, done.

PR #293, opened 2019-12-08
Speed up Firestore setup and make daily backups

janakdr:
We were unnecessarily reading prod data on every test case. Instead, download it once and write it each time. We were also (most likely unnecessarily) deleting test data before every test case. Since we have a timestamp and a random integer, forgo that.

This saves us ~1 second per test case.

Also write the downloaded data to a more "stable" location: the timestamp rounded to the nearest day, suffixed with "-backup". It is in our test/ collection, so will still be cleaned up by the delete-old-test-data functionality, but can be used before then. Along with this, extend the delete-old-test-data script tolerance to 7 days, so backups are more available.

Use Chrome on Travis, because it's more like what is used in prod, so why not. In my one Travis run with Chrome, it was 30 seconds faster than any previous Chromium run, although not a large sample.

Stop using Travis caching at all. The yarn and npm caches were useless, and took ~50 seconds to download. The .cache folder took ~73 seconds to download, but yarn can freshly install the packages in ~50 seconds.

Also not a scientific sample, but the Firestore change sped unit tests up by over 1 minute in my Travis run: https://dashboard.cypress.io/projects/jr8ks8/runs/1346/specs
image versus the old https://dashboard.cypress.io/projects/jr8ks8/runs/1344/specs
image

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -95,49 +95,60 @@ module.exports = (on, config) => {
             reject);
       });
     },
+
     /**
-     * Recursively deletes all data under the test/currentTestRoot tree, and
-     * creates necessary data for tests to consume, pulling that data from prod
-     * Firebase. For use before a test case runs. We add a dummy field inside
-     * test/currentTestRoot so that Firestore will deign to list this document
-     * later
-     * (https://stackoverflow.com/questions/47043651/this-document-does-not-exist-and-will-not-appear-in-queries-or-snapshots-but-id)
-     * @param {string} currentTestRoot document name directly underneath test/
-     * @return {Promise<null>} Promise that resolves when deletion and data
-     * creation is complete
+     * Retrieves necessary data for tests to consume from prod Firebase. Called
+     * once at the start of all tests.
+     * @return {Promise<Object>} Promise that contains all necessary disasters
+     *     data, as an array, with each element having a disaster key and a data
+     *     key.
      */
-    clearAndPopulateTestFirestoreData(currentTestRoot) {
-      const testAdminApp = createTestFirebaseAdminApp();
+    getTestFirestoreData() {
       const prodApp = firebase.initializeApp(firebaseConfigProd, 'prodapp');
+      const readPromises = [];
+      for (const disaster of ['2017-harvey', '2018-michael']) {
+        const documentPath = 'disaster-metadata/' + disaster;
+        const prodDisasterDoc = prodApp.firestore().doc(documentPath);
+        readPromises.push(
+            prodDisasterDoc.get()
+                .then((result) => ({disaster, data: result.data()})));
+      }
+      return Promise.all(readPromises).then(async (result) => {
+        await prodApp.delete();
+        return result;
+      })
+    },
+
+    /**
+     * Writes disasters data (retrieved using {@link getTestFirestoreData} into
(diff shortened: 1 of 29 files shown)


janakdr:
Reminder to merge this into correct branch.

cypress/plugins/index.js
```diff
@@ -6,6 +6,8 @@ const firebase = require('firebase');
 // https://on.cypress.io/plugins-guide
 // ***********************************************************
 
+let perTestFirestoreData;
```
juliexxia: Can you comment what this is

cypress/plugins/index.js
```diff
@@ -253,6 +242,33 @@ function deleteCollectionRecursively(collection) {
...
+  for (const disaster of ['2017-harvey', '2018-michael']) {
+    const documentPath = 'disaster-metadata/' + disaster;
+    const prodDisasterDoc = prodApp.firestore().doc(documentPath);
+    readPromises.push(prodDisasterDoc.get().then((result) => {
+      result.data().dummy = true;
+      return {disaster, data: result.data()};
+    }));
+  }
+  return Promise.all(readPromises)
+      .then(async (result) => {
```
juliexxia: Sorry, can you explain what async does here?

cypress/plugins/index.js
```diff
@@ -95,49 +98,34 @@ module.exports = (on, config) => {
...
-     * Recursively deletes all data under the test/currentTestRoot tree, and
-     * creates necessary data for tests to consume, pulling that data from prod
-     * Firebase. For use before a test case runs. We add a dummy field inside
-     * test/currentTestRoot so that Firestore will deign to list this document
-     * later
-     * (https://stackoverflow.com/questions/47043651/this-document-does-not-exist-and-will-not-appear-in-queries-or-snapshots-but-id)
-     * @param {string} currentTestRoot document name directly underneath test/
-     * @return {Promise<null>} Promise that resolves when deletion and data
-     * creation is complete
+     * Writes disasters data (retrieved using {@link storeTestFirestoreData}
```
juliexxia: Can you add a comment about when this is called (the one in storeTestFirestoreData already helpful)

cypress/plugins/index.js
```diff
@@ -253,6 +242,33 @@ function deleteCollectionRecursively(collection) {
   });
 }
 
+/**
+ * Stores necessary data for tests to consume from prod Firebase. Called
+ * once at the start of all tests. We
```
juliexxia: nit: Initially read this as all the tests in a single test file

cypress/plugins/index.js
janakdr: Done.

cypress/plugins/index.js
janakdr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

It allows us to put an await in here and then return the result of the promise without jumping through hoops.

cypress/plugins/index.js
janakdr: That's correct, clarified

cypress/plugins/index.js
janakdr: Done.

cypress/plugins/index.js
janakdr: Done.

PR #295, opened 2019-12-09
Validate values of select rather underlying data

janakdr:
This improves error messaging to users when the stored value is not displayed in the select (because, for instance, the user deleted an asset after previously selecting it here).

Note that if the user then re-creates the asset, they'll have to reload the page, but that seems ok.

We don't overwrite the value in Firestore, even though it's not valid, until the user changes it manually, to avoid the user losing data if an asset is momentarily not there (because they're deleting and recreating it).

### DIFF
--- a/cypress/integration/unit_tests/manage_disaster_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_test.js
@@ -323,10 +323,8 @@ describe('Unit tests for manage_disaster.js', () => {
           data.asset_data.damage_asset_path = 'pathnotfound';
           data.asset_data.snap_data.paths.NY = missingSnapPath;
           disasterData.set(getDisaster(), data);
-          Promise.all([
-            initializeScoreSelectors(['NY']),
-            initializeDamageSelector(['damage1', 'damage2']),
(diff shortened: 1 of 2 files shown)


cypress/integration/unit_tests/manage_disaster_test.js
```diff
@@ -314,6 +315,35 @@ describe('Unit tests for manage_disaster.js', () => {
...
 
+  it('nonexistent asset not ok', () => {
+    const missingSnapPath = 'whereisasset';
+    setUpAssetValidationTests()
+        .then(() => {
+          const data = createDisasterData(['NY']);
+          data.asset_data.damage_asset_path = 'pathnotfound';
+          data.asset_data.snap_data.paths.NY = missingSnapPath;
+          disasterData.set(getDisaster(), data);
+          Promise.all([
```
juliexxia: do you not need a return before this Promise.all?

janakdr: Great point. But actually these things don't return Promises anyway, so we can just execute them :)

PR #296, opened 2019-12-09
Support JSON tiles when defining map tile overlays

### LINKED ISSUE #273: Improve map tile import
Allow the json file to be used directly. Also accept lower-case z,x,y

janakdr:
Also accept lower-case '{z}', etc., when doing replacements, and replace throughout the string, in case there's some weird URL where you have to replace the same {z} multiple times.

Fixes #273


janakdr:
Reminder to merge into master once upstream is merged

docs/layer_util.js
```diff
@@ -324,20 +323,58 @@ function createTileCallback(layerDisplayData, resolve) {
...
   const layerDisplayData = new LayerDisplayData(null, true);
   layerArray[layer['index']] = layerDisplayData;
-  layerDisplayData.overlay = new CompositeImageMapType({
-    tileUrls: layer['urls'],
-    tileSize: layer['tile-size'],
-    maxZoom: layer['maxZoom'],
-    opacity: layer['opacity'],
+  const urlPromises = [];
+  for (const url of layer['urls']) {
+    if (url.match(/{Z}/i) && url.match(/{Y}/i) && url.match(/{X}/i)) {
```
juliexxia: can you do some commenting of regex expressions for the uninitiated

cypress/integration/unit_tests/checkbox_test.js
```diff
@@ -463,6 +465,69 @@ describe('Unit test for toggleLayerOn', () => {
...
+
+  it('tests slow json url with toggle off and on', () => {
+    let map = null;
+    createGoogleMap()
+        .then((returnedMap) => map = returnedMap)
+        .then(
+            () => map.fitBounds(new google.maps.LatLngBounds(
+                new google.maps.LatLng({lat: 26, lng: -97.4}),
+                new google.maps.LatLng({lat: 27.62, lng: -97.1}))));
+    let resolveFunction = null;
```
juliexxia: same as just let resolveFunction ?

janakdr: intellij on my mac complains for me sometimes if not initialized, because it doesn't know it gets set immediately in the Promise lambda. But seems ok now, so removed.

docs/layer_util.js
janakdr: Done, here and in CompositeImageMapType class.

Issue #270, opened 2019-12-04
In no-damage score asset creation make good defaults


juliexxia:
default snap to 1.0 weight and damage threshold to zero (and maybe also poverty threshold?)

janakdr:
Maybe also hide the "weights" panel entirely

PR #297, opened 2019-12-09
Good toggle story in the no-disaster-asset scenario

juliexxia:
When we don't have a disaster asset, don't show the weight or disaster threshold toggles. The weight is set to 100% poverty and the disaster threshold is set to 0.


cypress/integration/unit_tests/process_joined_data_test.js
```diff
@@ -20,48 +20,67 @@ joinedDataPromise.then = (lambda) => lambda({features: [feature]});
...
-    expect(returnedFeature).to.haveOwnProperty('properties');
-    const resultProperties = returnedFeature.properties;
-    // We modify the properties in place.
-    expect(resultProperties).to.equal(featureProperties);
-    expect(resultProperties).to.have.property('BLOCK GROUP', 'block group');
-    expect(resultProperties)
-        .to.have.property(
-            'SCORE', Math.round(100 * (0.5 * ((10 + 5) / 27) + 0.5 * (2 / 4))));
-    assertColorAndOpacity(resultProperties, 135);
+    cy.wrap(processJoinedData(
```
janakdr: Because you're not doing any Cypress commands after this, you can just chain the .then() off of the processJoinedData. In this test file and update_test.js as well.

docs/update.js
```diff
@@ -30,27 +29,66 @@ const damageWeightValueId = 'damage weight value';
...
+  if (!hasErrors(newValue, toggle)) {
+    toggles.set(toggle, newValue);
+  }
+}
+
+// Set in setUpInitialToggleValues.
+let hasDamageAsset = null;
+
+/**
+ * Initializes, damage-related toggle values based on whether or not we have
```
janakdr: nit: "Initializes,"?

docs/update.js
```diff
@@ -69,74 +107,79 @@ function createToggles(map) {
...
+  form.appendChild(createButton('update', () => update(map)));
   form.appendChild(document.createElement('br'));
   form.appendChild(createButton('current settings', reset));
 
   document.getElementById('form-div').appendChild(form);
   updateWeights();
 }
 
+/**
+ * Creates an div including input and label for the given toggle.
```
janakdr: nit: "an div"

docs/update.js
```diff
@@ -69,74 +107,79 @@ function createToggles(map) {
...
+ * @param {string} toggle
+ * @return {HTMLDivElement}
+ */
+function createInput(toggle) {
+  const thresholdInputDiv = document.createElement('div');
+  thresholdInputDiv.className = 'input-container';
+
+  const label = document.createElement('label');
+  label.for = toggle;
+  label.id = 'for-' + toggle;
```
janakdr: why do labels need ids?

docs/update.js
```diff
@@ -69,74 +107,79 @@ function createToggles(map) {
...
+  label.innerHTML = ' ' + toggle;
```
janakdr: why does this start with a space?

docs/run.js
```diff
@@ -34,12 +34,12 @@ const scalingFactor = 100;
...
   setMapToDrawLayersOn(map);
-  createToggles(map);
   snapAndDamagePromise =
       convertEeObjectToPromise(ee.FeatureCollection(snapAndDamageAsset));
-  createAndDisplayJoinedData(
-      map, initialPovertyThreshold, initialDamageThreshold,
-      initialPovertyWeight);
+  const initialTogglesValuesPromise =
+      setUpInitialToggleValues(disasterMetadataPromise);
+  initialTogglesValuesPromise.then(() => createToggles(map));
```
janakdr: why not have setUpInitialToggleValues do everything instead of passing things around like this?

juliexxia: Done

docs/update.js
juliexxia: Yeah, this whole file is kind of a mess (re: this comment and below). Wrote it in the early days. I just functionalized this code without changing it (I believe). I originally started to make it better but felt like it was obscuring all the actual changes. I can do the updates here if you're fine with that or add a TODO.  Removed the one label id for now but I think there's a bit more to be done here.

docs/update.js
juliexxia: Not sure. Removed. Probably some heinous attempt at buffering.

cypress/integration/unit_tests/update_test.js
```diff
@@ -1,100 +1,160 @@
...
-    document.getElementById('update').click();
+      slider.oninput();
 
-    expect(document.getElementById('snackbar-text').innerHTML)
-        .to.equals('poverty threshold must be between 0.00 and 1.00');
-    expect(document.getElementById('error').innerHTML)
-        .to.equals('ERROR: poverty threshold must be between 0.00 and 1.00');
-    expect(lastPassedPovertyThreshold).to.be.null;
+      expect(document.getElementById('poverty weight value').innerHTML)
+          .to.equals('0.01');
```
janakdr: super-nit, but "to.equal" is more grammatical :)

docs/update.js
```diff
@@ -30,27 +29,68 @@ const damageWeightValueId = 'damage weight value';
...
+function setUpToggles(disasterMetadataPromise, map) {
+  const togglesSetPromise = disasterMetadataPromise.then((doc) => {
+    hasDamageAsset = doc.data()['asset_data']['damage_asset_path'];
+    if (hasDamageAsset) {
+      toggles.set(damageThresholdKey, 0.5);
+      toggles.set(povertyWeightKey, 0.5);
+    }
+    return getValuesAsArray();
+  });
+  togglesSetPromise.then(() => createToggles(map));
```
janakdr: any reason not to do this in then block above? and then inline toggleSetPromise?

docs/update.js
```diff
@@ -30,27 +29,68 @@ const damageWeightValueId = 'damage weight value';
...
+  return togglesSetPromise;
+}
+
+/**
+ * Gets all the toggle values as an array.
+ * @return {[number, number, number]}
+ */
+function getValuesAsArray() {
```
janakdr: We're definitely not practicing this everywhere, but since this value is shared across files, what do you think of making it more structured? it can have povertyThreshold, etc, properties. That will also let you destructure it easily in the consumer.

docs/process_joined_data.js
```diff
@@ -52,24 +52,25 @@ function colorAndRate(
...
-      colorAndRate(
-          feature, scalingFactor, povertyThreshold, damageThreshold,
-          povertyWeight);
-    }
-    return featureCollection.features;
-  });
+    dataPromise, scalingFactor, initialTogglesValuesPromise) {
+  return Promise.all([dataPromise, initialTogglesValuesPromise])
+      .then((results) => {
+        const featureCollection = results[0];
```
janakdr: can destructure in one line, like `const [featureCollection, [povertyThreshold, ...]] = results;` if you take my suggestion below, just change inner square brackets to curlies.

docs/update.js
janakdr: Go as far as you want: github can't cope with indentation diffs anyway, so it's as if it were all new

docs/update.js
juliexxia: I didn't want other stuff chained off togglesSetPromise to have to wait on createToggles to finish, even if that's quick.

janakdr: As discussed offline, this is UI work, so should probably be prioritized, since the Promise consumer presumably isn't kicking off more async work based on this value.

juliexxia: Done.

docs/update.js
juliexxia: Done

cypress/integration/integration_tests/click_feature_test.js
```diff
@@ -22,15 +22,8 @@ describe('Integration test for clicking feature', () => {
 });
 
 /** Convenience function for clicking on the block group we use for testing. */
-function clickAndVerifyBlockGroup() {
+function clickBlockGroup() {
   cy.get('[placeholder="Search"]').clear().type('Greater Greenspoint{enter}');
-
   cy.wait(4000);
   cy.get('.map').click(900, 500);
-  cy.get('.map').should('contain', 'SCORE: 72');
```
janakdr: Can we leave a TODO for this? I feel nervous about completely losing coverage

juliexxia:
Closes  #270

PR #299, opened 2019-12-10
Use disaster picker for manage_layers.html

janakdr:
* Prettify disaster picker a bit, based on what manage_layers.html was doing: have a "Pending..." select until Firestore fetch completes.
* Allow title to be passed in to navbar even if we have a picker
* Give picker select a nice buffer so title is unlikely to jump around (it may still if there are very long disaster names)
* Get rid of title-only navbar, currently only used by dead add_asset.html
* Kick off manage_layers EE fetch for disaster assets even before Firestore fetch completes, and cache the Promise for the fetch, so that we can hide some EE latency behind Firestore latency.


docs/style.css
```diff
@@ -140,6 +140,14 @@ h1 {
   display: inline-block;;
   float: left;
   margin: 0;
+  min-width: 250px;
+}
+
+.nav-title {
+  display: inline-block;;
```
RamV13: Nit: unnecessary extra semicolon

docs/navbar.html
```diff
@@ -1,5 +1,6 @@
 <div class="navbar">
   <div id="nav-left" class="nav-left"></div>
+  <div id="nav-title" class="nav-title"><h1 class="nav-header" id="nav-title-header"></h1></div>
```
juliexxia: I don't think this id nav-title is used anywhere (mods made to class nav-title and id nav-title-header)

docs/import/manage_layers_startup.js
```diff
@@ -1,22 +1,24 @@
 import {Authenticator} from '../authenticate.js';
-import {loadNavbarWithTitle} from '../navbar.js';
+import {getDisastersData} from '../firestore_document.js';
+import {loadNavbarWithPicker} from '../navbar.js';
 import TaskAccumulator from '../task_accumulator.js';
+
```
juliexxia: nit: rm /n

PR #307, opened 2019-12-10
Disable disaster picker and notify user on Firestore saves

janakdr:
* Add a "toast" function that can give user updates. I tried to style it to be not too dreadful, but suggestions welcome. Before it fades out, looks like 
image
* Disable disaster picker while Firestore writes are in progress. This prevents a potential bug in which a user kicks off a write, makes another change, and then switches disasters, all before the write finishes. When the write finishes, it will see that a new write is necessary, but that new write will just write the *new* disaster's data, not the old disaster's data, which is effectively lost.
* Update tests to assert on new functionality, and in manage_disaster_test, get rid of the wait() hacks that we had.


janakdr:
Still to come: adding this functionality to the main map page, and adding the toast during initial page load.

RamV13:
Why don't we just reuse the snackbar added for error messaging? https://github.com/givedirectly/Google-Partnership/blob/master/docs/error.js#L27

janakdr:
I originally used that, but the snackbar looks much scarier, and is more prominent than I wanted these messages to be. Do you think we should make the size and color much more customizable and use it for both? Seemed easy enough to have them separate.

RamV13:
I was just thinking it would be nice to have a consistent look for these types of notifications. I can take care of it in a follow up PR.

janakdr:
What would make that consistent look easier for you to do? Would you like me to put these in the snackbar for now? It just looked very scary. I can also have the toast show up at the bottom. Let me know!

docs/import/manage_disaster.html
```diff
@@ -36,6 +36,8 @@
 <div id="navbar"></div>
 <div class="wrapper">
 <div class="main-content">
+  <div id="toast-message"><span id="toast-message-text"></span></div>
+
```
juliexxia: nit: extra line

RamV13:
Putting it in the snackbar would be nice but it's okay if it's too much trouble!

PR #308, opened 2019-12-11
Be smarter about parsing number values during score asset creation

### LINKED ISSUE #269: Median income max min computed on strings so e.g. 100 < 50
The larger fix here is not computing max/min for string values and allowing the color function to handle null values for cases like median income where there are some missing values.

juliexxia:
Run SVI, median income, snap and total pop values through a parser while creating the score asset. The parser takes care of values like "250,000+" (250000) and non-numerics (null). Also make null values transparent for layer creation. 

closes #269

### DIFF
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -62,7 +62,7 @@ function createContinuousFunction(field, minVal, maxVal, color) {
   const range = maxVal - minVal;
   return (feature) => {
     let value = feature['properties'][field];
-    if (!value) return transparent;
+    if (value === null) return transparent;
     if (value > maxVal) {
       value = maxVal;
     } else if (value < minVal) {
@@ -90,7 +90,7 @@ function createContinuousFunction(field, minVal, maxVal, color) {
 function createDiscreteFunction(field, colors) {
   return (feature) => {
     const value = feature['properties'][field];
-    if (!value) return transparent;
+    if (value === null) return transparent;
     const rgba = colorMap.get(colors[value]);
     rgba.push(opacity);
     return rgba;
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -62,6 +62,7 @@ function createContinuousFunction(field, minVal, maxVal, color) {
   const range = maxVal - minVal;
   return (feature) => {
     let value = feature['properties'][field];
+    if (!value) return transparent;
     if (value > maxVal) {
       value = maxVal;
     } else if (value < minVal) {
@@ -88,12 +89,15 @@ function createContinuousFunction(field, minVal, maxVal, color) {
  */
 function createDiscreteFunction(field, colors) {
   return (feature) => {
-    const color = colors[feature['properties'][field]];
-    const rgba = colorMap.get(color);
+    const value = feature['properties'][field];
+    if (!value) return [0, 0, 0, 0];
+    const rgba = colorMap.get(colors[value]);
     rgba.push(opacity);
     return rgba;
   };
 }
+
+const transparent = [0, 0, 0, 0];
 
 const colorMap = new Map([
   ['red', [255, 0, 0]],
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -90,7 +90,7 @@ function createContinuousFunction(field, minVal, maxVal, color) {
 function createDiscreteFunction(field, colors) {
   return (feature) => {
     const value = feature['properties'][field];
-    if (!value) return [0, 0, 0, 0];
+    if (!value) return transparent;
     const rgba = colorMap.get(colors[value]);
     rgba.push(opacity);
     return rgba;
--- a/docs/import/create_score_asset.js
+++ b/docs/import/create_score_asset.js
@@ -57,7 +57,7 @@ function countDamageAndBuildings(feature, damage, buildings) {
 
 // Permissive number regexp: matches optional +/- followed by 0 or more digits
 // followed by optional period with 0 or more digits. Corresponds to valid
-// inputs to ee.Number.parse. As with ee.Number.parse, the empty string is not
+// inputs to ee.Number.parse. Like ee.Number.parse, the empty string is not
 // allowed.
 const numberRegexp = '^[+-]?(([0-9]*)?[0-9](\.[0-9]*)?|\.[0-9]+)$';
 
@@ -66,7 +66,7 @@ const endsWithPlusMinusRegexp = '[+-]$';
 
 /**
  * First strips out all ',' and whitespace. Then strips any trailing '-' or '+'
- * (for threshold values like '250,000+'). Then checks if the result is matches
+ * (for threshold values like '250,000+'). Then checks if the result matches
  * our regexp for valid EE numbers. If it does, parse and return number, else
  * return null.
  * @param {string} value
--- a/docs/import/create_score_asset.js
+++ b/docs/import/create_score_asset.js
@@ -1,5 +1,6 @@
 import {blockGroupTag, buildingCountTag, damageTag, geoidTag, incomeTag, snapPercentageTag, snapPopTag, sviTag, totalPopTag, tractTag} from '../property_names.js';
 import {getScoreAsset} from '../resources.js';
+
 import {computeAndSaveBounds, saveBounds} from './center.js';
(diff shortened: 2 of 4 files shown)


docs/import/create_score_asset.js
```diff
@@ -22,16 +22,20 @@ function countDamageAndBuildings(feature, damage, buildings) {
   const geoId = feature.get(geoidTag);
   const totalBuildings = ee.Algorithms.If(
       buildings.contains(geoId), buildings.get(geoId), ee.Number(0));
+  const rawIncomeValue = ee.String(feature.get(incomeTag));
+  const income = ee.Algorithms.If(
+      ee.Algorithms.IsEqual(rawIncomeValue, ee.String('250,000+')),
+      ee.Number(250000),
+      ee.Algorithms.If(
+          ee.Algorithms.IsEqual(rawIncomeValue, ee.String('-')), null,
+          ee.Number.parse(rawIncomeValue)));
```
janakdr: should we just filter out all non-numeric values? Are we sure these are only ones? We could do for svi too as a safety measure (other values are known numbers). For example, here is functionality to strip out all non-numeric values:

```
// Permissive number regexp: matches optional whitespace followed by optional
// minus sign followed by 0 or more digits followed by optional period with 0 or
// more digits followed by optional whitespace. Note that it matches an empty
// string. Besides empty string, corresponds to valid inputs to ee.Number.parse.
const numberRegexp = '^ *-?[0-9]*(\.[0-9]*) *$';

function convertToNumber(value) {
  value = ee.String(value);
  return ee.Algorithms.If(
      value.length().and(value.match(numberRegexp).length()),
      ee.Number.parse(value),
      null);
}

```

docs/import/create_score_asset.js
```diff
@@ -22,16 +22,20 @@ function countDamageAndBuildings(feature, damage, buildings) {
   const geoId = feature.get(geoidTag);
   const totalBuildings = ee.Algorithms.If(
       buildings.contains(geoId), buildings.get(geoId), ee.Number(0));
+  const rawIncomeValue = ee.String(feature.get(incomeTag));
+  const income = ee.Algorithms.If(
+      ee.Algorithms.IsEqual(rawIncomeValue, ee.String('250,000+')),
```
janakdr: maybe leave a TODO to handle any '<large number>+' similarly? I can imagine Census changes threshold

docs/import/create_score_asset.js
```diff
@@ -22,16 +22,20 @@ function countDamageAndBuildings(feature, damage, buildings) {
...
+      ee.Number(250000),
+      ee.Algorithms.If(
+          ee.Algorithms.IsEqual(rawIncomeValue, ee.String('-')), null,
```
janakdr: how do we handle null values when rendering a layer? Is that a follow-up? Also, how do we handle it in pop-up display?

docs/import/create_score_asset.js
juliexxia: Done. With updated comments.

docs/import/create_score_asset.js
juliexxia: Before it was just rendering as white since null < all numbers but now we're setting it to be transparent if the val is null. Probably a check to have to prevent referencing null in the discrete case as well (since the color map is value -> color).

docs/import/create_score_asset.js
juliexxia: Done in this PR

docs/firebase_layers.js
```diff
@@ -62,6 +62,7 @@ function createContinuousFunction(field, minVal, maxVal, color) {
   const range = maxVal - minVal;
   return (feature) => {
     let value = feature['properties'][field];
+    if (!value) return transparent;
```
janakdr: should probably check explicitly for null. What if we have a range in which 0 is valid? Same for below

docs/import/create_score_asset.js
```diff
@@ -49,7 +39,41 @@ function countDamageAndBuildings(feature, damage, buildings) {
   } else {
     properties = properties.set(damageTag, 0);
   }
-  return ee.Feature(geometry, properties);
+  const snapPop = ee.Number(feature.get(snapPopTag)).long();
```
janakdr: This works if feature.get(snapPopTag) is null? I would expect it to throw an error.

docs/import/create_score_asset.js
```diff
@@ -49,7 +39,41 @@ function countDamageAndBuildings(feature, damage, buildings) {
...
+      .set(snapPercentageTag, snapPop.divide(totalPop));
+}
+
+// Permissive number regexp: matches optional +/- followed by 0 or more digits
+// followed by optional period with 0 or more digits. Corresponds to valid
+// inputs to ee.Number.parse.
+const numberRegexp = '^[+-]?(([0-9]*)?[0-9](\.[0-9]*)?|\.[0-9]+)$';
+
+// Regexp that matches any string that ends in a '+' or '-'.
+const endsWithPlusMinusRegexp = '.*[+-]$';
```
janakdr: nit: you can leave off the '.*' since you're not anchoring the regexp to the start of the expression (with '^')

docs/import/create_score_asset.js
```diff
@@ -49,7 +39,41 @@ function countDamageAndBuildings(feature, damage, buildings) {
...
+  // convertToNumber} so must be set directly on feature, not in dictionary.
+  return ee.Feature(geometry, properties)
+      .set(incomeTag, feature.get(incomeTag))
+      .set(sviTag, feature.get(sviTag))
+      .set(snapPopTag, snapPop)
+      .set(totalPopTag, totalPop)
+      .set(snapPercentageTag, snapPop.divide(totalPop));
+}
+
+// Permissive number regexp: matches optional +/- followed by 0 or more digits
```
janakdr: Might want to clarify that empty string is not allowed.

docs/import/create_score_asset.js
```diff
@@ -49,7 +39,41 @@ function countDamageAndBuildings(feature, damage, buildings) {
...
+// followed by optional period with 0 or more digits. Corresponds to valid
+// inputs to ee.Number.parse.
+const numberRegexp = '^[+-]?(([0-9]*)?[0-9](\.[0-9]*)?|\.[0-9]+)$';
+
+// Regexp that matches any string that ends in a '+' or '-'.
+const endsWithPlusMinusRegexp = '.*[+-]$';
+
+/**
+ * First strips out all ',' and whitespace. Then strips any trailing '-' or '+'
+ * (for threshold values like '250,000+'). Then checks if the result is parsable
```
janakdr: nit: I might say, "...if the result matches our regexp for valid EE numbers". The phrasing almost makes me think we're trying to parse as a number and catching errors.

docs/firebase_layers.js
juliexxia: Ah good point

docs/import/create_score_asset.js
juliexxia: Didn't see an algs 'or' function. Figure this works? If either is null we have no real shot at getting a percentage.

docs/firebase_layers.js
```diff
@@ -88,13 +89,16 @@ function createContinuousFunction(field, minVal, maxVal, color) {
  */
 function createDiscreteFunction(field, colors) {
   return (feature) => {
-    const color = colors[feature['properties'][field]];
-    const rgba = colorMap.get(color);
+    const value = feature['properties'][field];
+    if (value === null) return transparent;
```
janakdr: I think this is fine, but just FYI, I think you could even just add [null, transparent] as an entry to the map. Javascript apparently allows null keys.

docs/import/create_score_asset.js
```diff
@@ -49,7 +39,47 @@ function countDamageAndBuildings(feature, damage, buildings) {
   } else {
     properties = properties.set(damageTag, 0);
   }
-  return ee.Feature(geometry, properties);
+  const snapPop = feature.get(snapPopTag);
+  const totalPop = feature.get(totalPopTag);
+  // unfortunately, just using .contains(null) here throws an error.
+  const snapPercentage = ee.Algorithms.If(
```
janakdr: This works, but might be slightly more straightforward to do `ee.Algorithms.If(snapPop, ee.Algorithms.If(totalPop, ee.Number(snapPop).long().divide(ee.Number(totalPop).long()), null), null)`. Up to you.

janakdr:
Maybe change PR description

Issue #141, opened 2019-10-17
Legend for the layers


juliexxia:
- what color the layer is
- description of the layer (pops up on check?)

janakdr:
I don't think we have any available descriptions of the layers beyond the already shown display name. And I also think that showing technical info like EE asset name is probably not what a user wants when mousing over a working layer. So declaring this fixed.

PR #309, opened 2019-12-11
Add a legend to the layers sidebar

ruthtalbot:
Lmk what you think -- it feels a little busy


juliexxia:
Hmm, I think you're right. Do you know why the FEMA Assistance one and the Damage one look different? They should be the same schema right (discrete, two colors?)

Maybe the coloring checkboxes is the way to go?

docs/run.js
```diff
@@ -97,29 +97,61 @@ function drawTableAndSetUpHandlers(processedData, map, scoreAsset) {
...
+    return '';
+  }
+  const colors = [];
+  const currentStyle = colorFunction['current-style'];
+  switch (currentStyle) {
+    case 0:
+      colors.push('white', colorFunction['color']);
+      break;
+    case 1:
+      colors.push(...(new Set(Object.values(colorFunction['colors']))));
```
janakdr: nit: any reason to spread out a Set rather than an Array? Should be a bit more efficient.

docs/run.js
```diff
@@ -97,29 +97,61 @@ function drawTableAndSetUpHandlers(processedData, map, scoreAsset) {
...
+ */
+function getLinearGradient(colorFunction) {
+  if (!colorFunction) {
+    return '';
+  }
+  const colors = [];
+  const currentStyle = colorFunction['current-style'];
+  switch (currentStyle) {
+    case 0:
+      colors.push('white', colorFunction['color']);
```
janakdr: I guess this is so that we can tell the difference between single-color and continuous? I'm not sure it's adding value, though, especially since it's then unclear on the difference between discrete and continuous. If you do want to distinguish, I'd suggest doing discrete vertically and continuous horizontally.

docs/run.js
```diff
@@ -97,29 +97,61 @@ function drawTableAndSetUpHandlers(processedData, map, scoreAsset) {
...
  * @param {String} displayName checkbox display name
  * @param {div} parentDiv div to attach checkbox to
+ * @param {Object} colorFunction color data from the layer
  * @return {HTMLInputElement} the checkbox
  */
 function createNewCheckbox(index, displayName, parentDiv) {
   const newRow = document.createElement('div');
   newRow.className = 'checkbox-row';
+
   const newBox = document.createElement('input');
```
janakdr: A nice feature (follow-up?) would be if you moused over the layer name, it gave you further information. For the discrete case, that would be really useful, since it could tell you what each color meant. For continuous it could give range. Could also say column that was being displayed.

docs/run.js
ruthtalbot: Not sure how common this will be, but GD visits has only 3 possible colors, but they're reused repeatedly, so I wanted to filter it down to only unique colors so it didn't try to create a gradient between red yellow red red orange or something like that.

janakdr: Oh, great point on deduping. Carry on :)

docs/run.js
ruthtalbot: Updated -- makes the code uglier but is much clearer

docs/run.js
```diff
@@ -132,29 +132,62 @@ function drawTableAndSetUpHandlers(processedData, map) {
...
+    return '';
+  }
+  const currentStyle = colorFunction['current-style'];
+  let gradientString = 'linear-gradient(to right';
+  switch (currentStyle) {
+    case 0:
+      gradientString += ', white,' + colorFunction['color'];
+      break;
+    case 1:
+      const colors = [...(new Set(Object.values(colorFunction['colors'])))];
```
janakdr: is it possible for colors to be empty?

docs/run.js
```diff
@@ -132,29 +132,62 @@ function drawTableAndSetUpHandlers(processedData, map) {
...
   return newBox;
 }
 
+/**
+ * Gets the linear gradient of the colors for the legend.
+ *
+ * @param {Object} colorFunction color data from the layer
+ * @return {string} the linear gradient
+ */
+function getLinearGradient(colorFunction) {
```
janakdr: might be worth unit-testing this.

docs/run.js
janakdr: Add a TODO for this feature?

docs/run.js
```diff
@@ -166,10 +199,12 @@ function createNewCheckbox(index, displayName, parentDiv) {
 function createNewCheckboxForLayer(layer, parentDiv, map) {
   const index = layer['index'];
   const newBox = createNewCheckbox(index, layer['display-name'], parentDiv);
-  if (!layer['display-on-load']) {
-    newBox.checked = false;
-  }
+  const linearGradient = getLinearGradient(layer['color-function']);
+  newBox.checked = layer['display-on-load'];
```
janakdr: do `!!layer['display-on-load']` to make sure it's just a boolean?

docs/run.js
ruthtalbot: I'd assume so -- I think worst case if the linear gradient is invalid it just defaults back to the green checkbox though. I'll double check

docs/run.js
ruthtalbot: Added!

docs/import/color_function_util.js
```diff
@@ -458,3 +458,35 @@ function getColorFunction() {
...
+  }
+  const currentStyle = colorFunction['current-style'];
+  let gradientString = 'linear-gradient(to right';
+  switch (currentStyle) {
+    case 0:
+      gradientString += ', white, ' + colorFunction['color'];
+      break;
+    case 1:
+      const colors = [...(new Set(Object.values(colorFunction['colors'])))];
+      const percent = 100 / colors.length;
```
janakdr: In the case that you have a discrete layer but have chosen no colors for it, this results in percent = Infinity. Not awesome, although code tolerates it. Should probably add a check for this.

janakdr:
This did a lot of the work needed to close #141

PR #311, opened 2019-12-12
Consolidating toasts and errors into the snackbar

RamV13:


docs/snackbar.js
```diff
@@ -0,0 +1,57 @@
...
+  const snackbarOriginalClasses = snackbar.className;
+  const snackbarOriginalStyle = snackbar.style;
+  if (snackbarStyler) snackbarStyler(snackbar);
+
+  const icon = document.getElementById('snackbar-icon');
+  const iconOriginalClasses = icon.className;
+  const iconOriginalStyle = icon.style;
+  if (iconStyler) iconStyler(icon);
+
+  document.getElementById('snackbar-text').innerHTML = msg;
```
janakdr: Maybe innerText?

docs/snackbar.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} msg the message to be shown
+ * @param {function} snackbarStyler applies styles to the snackbar div
+ * @param {function} iconStyler applies styles to the icon div
+ * @param {number} duration Number of milliseconds to show the message. If
+ *     non-positive, message stays up until this method is called again.
+ */
+function showSnackbar(
+    msg, snackbarStyler, iconStyler, duration = SNACKBAR_DURATION_MS) {
+  const snackbar = document.getElementById('snackbar');
+  const snackbarOriginalClasses = snackbar.className;
```
janakdr: I think that if two messages come in quickly, the second message will restore the first message's style, which is not what we want, right? Maybe just create new elements each time?

RamV13: Ah thanks for catching that!

I think adding more elements makes it more complicated so I just initialized the original style on load.

docs/snackbar.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {string} msg the message to be shown
+ * @param {function} snackbarStyler applies styles to the snackbar div
+ * @param {function} iconStyler applies styles to the icon div
+ * @param {number} duration Number of milliseconds to show the message. If
+ *     non-positive, message stays up until this method is called again.
+ */
+function showSnackbar(
+    msg, snackbarStyler, iconStyler, duration = SNACKBAR_DURATION_MS) {
+  const snackbar = document.getElementById('snackbar');
+  if (snackbarStyler) snackbarStyler(snackbar);
```
janakdr: I think you're only setting the snackbar element's style. Might be easier to just pass the styles in as a dict and the classnames in as a list and then apply it yourself in here. That also guarantees that the lambda didn't do anything else to the element, since you're only resetting the classes and style. Same goes for icon. Maybe we should only allow setting class names, and then the toast and the error styles can be classes?

RamV13: Fair enough

janakdr:
As a simple workaround for the Cypress errors, maybe just create the html elements in JavaScript rather than loading a page?

RamV13:
Yup I went that route but unfortunately `click_feature_test.js` seems to be broken at head even with #313 so the build is still failing.

docs/snackbar.js
```diff
@@ -0,0 +1,77 @@
...
+
+  if (duration > 0) {
+    setTimeout(() => {
+      // Only clear the snackbar if the current shown snackbar is the latest.
+      if (currentShow === showNumber) {
+        showNumber = 0;  // reset the counter to prevent overflow
+        $('#snackbar').fadeOut(ANIMATION_DURATION_MS, () => {
+          // In case another snackbar is already being shown again, we don't
+          // want to reset the styles.
+          if (showNumber > 0) return;
```
janakdr: What happens if fadeIn is called while fadeOut is in progress? Does it get totally canceled, and the callback isn't called? If so, then this guard isn't necessary, right?

If the callback is called even if another message has faded in, then I think this is a bug. What if message1 comes in, timeout hits, it starts fading out, then message2 comes in, timeout hits, it sets showNumber to 0, then starts fading out, then message1's fadeOut callback hits, and resets the style before message2's fadeOut finishes. I would suggest resetting the showNumber inside the fadeOut callback, or not even resetting it (9007199254740991 messages ought to be enough for anyone).

RamV13: I think we need to keep the `showNumber` reset because that's how we identify whether or not there is a snackbar already being shown again and bail out of resetting the styles.

Moved the reset to within the fade out callback just in case.

janakdr: I don't understand this now. You're not setting showNumber to 0 unless you get past this guard, so when will you ever pass this guard?

Can't you just check here that currentShow === showNumber like above? And not reset?

RamV13: Oops ... silly mistake.

Yup that works better.

PR #313, opened 2019-12-12
Fix bug in clicking feature when feature not in list

janakdr:
Oops!


juliexxia:
@janakdr seeing flakes in this new test on my current PR. Have you seen any?

juliexxia:
Fixed by moving around asserts like you did in the no svi/income PR

janakdr:
Yeah, I snuck that fix in. It's because the map.data add actually happens
on a different thread, and it isn't always triggered before we go look for
the feature in map.data. When we look for the selected row in the table, we
don't find it immediately and that gives the other thread time to run, then
we can proceed. Sorry about that!

PR #310, opened 2019-12-11
Column and existence verification for score assets

juliexxia:
On load and on asset picker change, verify the feature collection specified to go into the score asset (1) exists and (2) has the column headers we expect for that type of asset. 

This moves us more towards being very locked down on census column headers. 

#279

PR #314, opened 2019-12-13
Create score asset even if SVI/income missing

janakdr:
Tolerate absence of these assets, on a per-state basis.

Does not actually allow user to kick off processing without these assets: will have to think a bit more about how best to message that, probably using functionality being added in #310.

Also does not do anything for actual map display, but I verified that nothing crashes in prod, and also added a test here. "SVI: undefined" isn't the prettiest message, but is also reasonable. In future work, we could suppress unspecified columns, but it might also be a useful reminder, and clearer than someone wondering why SVI just isn't there.


cypress/integration/unit_tests/create_score_asset_test.js
```diff
@@ -162,6 +162,25 @@ describe('Unit tests for create_score_asset.js', () => {
...
+    expect(promise).to.not.be.null;
+    cy.wrap(promise)
+        .then(() => {
+          expect(exportStub).to.be.calledOnce;
+          return convertEeObjectToPromise(exportStub.firstCall.args[0]);
+        })
+        .then((result) => {
+          const features = result.features;
+          expect(features).to.have.length(1);
+          expect(features[0]['properties']['MEDIAN INCOME']).to.be.undefined;
```
juliexxia: check that other things are defined?

janakdr: Nothing else is defined, because this feature doesn't have any properties (besides score) set at all, because it wasn't created by the score code, just created manually. I think it's fine to leave as is: if we get stricter about requiring other fields we can add them to the test then.

janakdr:
Changed slightly to not tolerate null state dictionaries, since that should never happen and it adds a bit of code complexity.

Issue #283, opened 2019-12-05
Make bounds selection easier (polygon drawn on map)


janakdr:

janakdr:
Main remaining work here is better documentation and showing layers on this map.

janakdr:
Giving to Ruth because it sounds like showing layers is a lower priority than documentation and a bunch of other things. After that's done, if we still have time to show layers, please assign back to me.

ruthtalbot:
Agreed, will hold on to it for now

PR #316, opened 2019-12-16
Zoom to disaster states in map when polygon coordinates absent

janakdr:
Use EE TIGER states resource. Not instant, so we don't zoom if user manages to beat us to a zoom.

It might have been nice to cancel our operation if user panned as well, but I don't know how to get that event: bounds_changed fires too often. Also, if the user is panning around the map but it's zoomed out, probably not much value in it.

There is a weird quirk right now: if you switch between two disasters, and the second one doesn't have a polygon, we won't re-center the map until the state ee processing is finished. Until then, the previous disaster's map will stay up (though without the polygon). We could re-center the map on the US until the ee promise finished, but that felt pretty manic.

#283

### DIFF
--- a/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
@@ -54,7 +54,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
                      .to.eql('asset2'));
   });
 
-  it.only('no map coordinates to start', () => {
+  it('no map coordinates to start', () => {
     const data = setUpDefaultData();
     data.asset_data.score_bounds_coordinates = null;
     callEnableWhenReady(data);
--- a/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
@@ -54,7 +54,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
                      .to.eql('asset2'));
   });
 
-  it('no map coordinates to start', () => {
+  it.only('no map coordinates to start', () => {
     const data = setUpDefaultData();
     data.asset_data.score_bounds_coordinates = null;
     callEnableWhenReady(data);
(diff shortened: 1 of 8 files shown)


docs/import/score_bounds_map.js
```diff
@@ -105,21 +114,39 @@ class ScoreBoundsMap {
...
+      google.maps.event.addListenerOnce(
+          this.map, 'zoom_changed', () => zoomChangedByUser = true);
+      return this.stateBoundsPromise.then((resolvedBounds) => {
+        if (zoomChangedByUser) {
+          return;
+        }
+        const coordinates = resolvedBounds.coordinates[0];
+        bounds.extend(latLngLiteralFromArray(coordinates[0]));
+        bounds.extend(latLngLiteralFromArray(coordinates[2]));
+        console.log('changed bounds inside');
```
juliexxia: debugging?

docs/import/score_bounds_map.js
```diff
@@ -105,21 +114,39 @@ class ScoreBoundsMap {
...
-      this.map.setZoom(defaultZoomLevel);
+      let zoomChangedByUser = false;
+      google.maps.event.addListenerOnce(
+          this.map, 'zoom_changed', () => zoomChangedByUser = true);
+      return this.stateBoundsPromise.then((resolvedBounds) => {
+        if (zoomChangedByUser) {
+          return;
+        }
+        const coordinates = resolvedBounds.coordinates[0];
+        bounds.extend(latLngLiteralFromArray(coordinates[0]));
```
juliexxia: can you document the significance of the 0 and 2 indices?

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -62,6 +61,20 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
     cy.get('#damage-asset-select').should('have.value', '');
     cy.get('#map-bounds-div').should('be.visible');
     cy.get('.score-bounds-delete-button').should('not.be.visible');
+    // Has NY in view after EE promise finishes and zoom happens.
+    cy.wrap(scoreBoundsMap.stateBoundsPromise).then(() => {
+      new Promise(
+          (resolve) => google.maps.event.addListenerOnce(
+              scoreBoundsMap.map, 'zoom_changed', () => {
+                expect(scoreBoundsMap.map.getBounds().contains({
+                  lng: -74,
```
juliexxia: Maybe add that these came from inspecting NY bounds (assuming that's how it happened?)

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -62,6 +61,20 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
...
+                  lat: 41.7,
+                })).to.be.true;
+                expect(underTest.map.getBounds().contains({lng: -100, lat: 32}))
```
juliexxia: And add what these points are supposed to test?

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -62,6 +61,20 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
     cy.get('#damage-asset-select').should('have.value', '');
     cy.get('#map-bounds-div').should('be.visible');
     cy.get('.score-bounds-delete-button').should('not.be.visible');
+    // Has NY in view after EE promise finishes and zoom happens.
+    cy.wrap(scoreBoundsMap.stateBoundsPromise).then(() => {
+      new Promise(
```
juliexxia: Why is creating and resolving this promise neccesary here?

janakdr: Added a nice long comment.

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
janakdr: Comment added.

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
janakdr: Done.

docs/import/score_bounds_map.js
janakdr: extracted some variables, hopefully the names help.

janakdr:
Messed up the test a bit: zoom_bounds listener actually does need to be registered immediately, it's too late when state promise already finished. I think flow is simpler now anyway. Will submit when Travis passed, but TAL if you like.

PR #317, opened 2019-12-16
UI clean up: Update table display/add headers/add space between sections

ruthtalbot:
Screen Shot 2019-12-16 at 10 32 21 AM
Screen Shot 2019-12-16 at 10 32 09 AM


docs/import/manage_disaster.html
```diff
@@ -103,19 +103,19 @@
     </form>
   </div>
 <hr/>
-  <div id="current-disaster-interaction">
-    <button type="button" id="delete" disabled>Delete Current Disaster</button>
-    <div id="score-asset-selection">
-      <div class="form-title">Select Score Data</div>
-      <div>
+  <div id="current-disaster-interaction" style="width:100%">
```
RamV13: I think it would be nice to keep all our styles in the CSS but if we're time pressed that's okay I can clean it up later.

docs/draw_table.js
```diff
@@ -76,11 +76,20 @@ function renderTable(list, features, map, selectorReceiver) {
   const dataView = new google.visualization.DataView(data);
   // don't display geoid
   dataView.hideColumns([0]);
+  //'headerRow': 'italic-darkblue-font large-font bold-font',
   const table =
       new google.visualization.Table(document.getElementById('table'));
-  table.draw(
-      dataView,
-      {page: 'enable', pageSize: 25, sortColumn: 1, sortAscending: false});
+  const classNames = {
```
janakdr: nit: is this worth keeping as a separate variable, as opposed to inlining?

docs/script.js
```diff
@@ -26,6 +26,6 @@ google.charts.load('current', {packages: ['table', 'controls']});
 $(() => {
   initializeSidebar();
   map = createMap(disasterMetadataPromise);
-  loadNavbarWithPicker(firebaseAuthPromise, 'GiveDirectly damage/poverty map');
+  loadNavbarWithPicker(firebaseAuthPromise, 'Delphi Map');
```
janakdr: Should also change https://github.com/givedirectly/Google-Partnership/blob/master/docs/index.html#L4

docs/navbar.html
```diff
@@ -1,7 +1,7 @@
 <div class="navbar">
-  <div id="nav-left" class="nav-left"></div>
   <div class="nav-title"><h1 class="nav-header" id="nav-title-header"></h1></div>
-
+  <div id="nav-left" class="nav-left"></div>
```
janakdr: Is this what we want? I feel like the disaster picker should be less prominent in general, since it's usually the same for weeks/months, but this feels like it's more prominent. Can you say what you want to accomplish with this change?

ruthtalbot: The title never changes, so it made sense to me to make the picker slightly more prominent. Visually it also read more intuitively to me -- like "Delphi Map of harvey". Also there could be a lot of extra space in the picker which sometimes looked strange when it was pushing the header out.

PR #318, opened 2019-12-16
Make SVI/Income/Buildings optional when possible

janakdr:
SVI and Income are always optional, Buildings optional when damage asset not present. Messaging stays in the button. We change the button color to a grayish yellow when missing optional assets.

### DIFF
--- a/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
@@ -80,7 +80,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
       ', and must specify either damage asset or map bounds' +
       allOptionalMissing;
 
-  it.only('validates asset data', () => {
+  it('validates asset data', () => {
     const boundsChanged = new Promise((resolve) => {
       const listener = scoreBoundsMap.map.addListener('bounds_changed', () => {
         google.maps.event.removeListener(listener);
@@ -161,6 +161,14 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
     cy.get('#process-button')
         .should('have.css', 'background-color')
         .and('eq', 'rgb(150, 150, 0)');
+    // Get rid of score asset.
+    setFirstSelectInScoreRow(0).select('').blur();
+    cy.get('#process-button')
+        .should('be.disabled')
+        .should('have.css', 'background-color')
+        .and('eq', 'rgb(128, 128, 128)');
+    // Put score asset back.
+    setFirstSelectInScoreRow(0);
     // Put income back.
     setFirstSelectInScoreRow(1);
     cy.get('#process-button')
@@ -176,7 +184,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
         .should('have.text', 'Must specify either damage asset or map bounds');
     cy.get('#process-button').should('be.disabled');
     // Get rid of score asset.
-    setFirstSelectInScoreRow(0).select('');
+    setFirstSelectInScoreRow(0).select('').blur();
     cy.get('#process-button')
         .should(
             'have.text',
--- a/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
@@ -147,10 +147,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
     }
     // Yay! We're ready to go.
     cy.get('#process-button')
-        .should(
-            'have.text',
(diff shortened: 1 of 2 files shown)


docs/import/manage_disaster.js
```diff
@@ -61,48 +61,90 @@ let scoreBoundsMap;
...
- * all validation succeeds, enables kick-off button, otherwise disables and
- * changes button text to say what is missing.
+ * Checks that all mandatory fields that can be entered by the user have
+ * non-empty values. Does not check that assets actually exist, are of valid
+ * type, etc. If all validation succeeds, enables kick-off button, otherwise
+ * disables and changes button text to say what is missing.
+ *
+ * Some fields (Income, SVI) are optional. If they are missing, a separate
+ * message is displayed on the button, but it can still be enabled. If it is
+ * enabled, it is yellowed a bit to indicate the missing optional assets.
```
juliexxia: So the color of the button for unable to kick off and able but missing is the same? What would you think about making them different? I think my slight preference is for anytime the button can actually be used for kicking off, it's green. But @ruthtalbot what do you think?

docs/import/manage_disaster.js
```diff
@@ -61,48 +61,90 @@ let scoreBoundsMap;
...
+ * The buildings asset is optional if the damage asset is not present, but is
+ * required if the damage asset is present.
  */
 function validateUserFields() {
   const states = disasterData.get(getDisaster()).states;
   const missingItems = [];
-  for (const scoreAssetType of scoreAssetTypes) {
+  const damageAssetPresent = !!$('#damage-asset-select').val();
```
juliexxia: nit: would you mind also doing  this in update.js for the initialization of the hasDamageAset var?

juliexxia: can you document what the missingItems array ends up holding before we start processing it?

docs/import/manage_disaster.js
```diff
@@ -61,48 +61,90 @@ let scoreBoundsMap;
...
-  const hasDamage = $('#damage-asset-select').val() ||
-      getElementFromPath(scoreCoordinatesPath);
+  const hasDamage =
+      damageAssetPresent || getElementFromPath(scoreCoordinatesPath);
   let message = '';
+  let optionalMessage = '';
   if (missingItems.length) {
-    message = 'Missing asset(s): ' + missingItems.join(', ');
+    for (const missingItem of missingItems) {
+      const optionalBuildings =
```
juliexxia: This logic is starting to get quite hairy (though looks good in the button!) If we have time (lol) maybe we can think about more visual messaging through icons or something.

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -69,13 +69,18 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
...
+  const alwaysOptionalMissing =
+      '; warning: created asset will be missing Income, SVI';
+  const allOptionalMissing = alwaysOptionalMissing + ', Building counts';
+  const allStateAssetsMissingWithDamageAssetText = allMandatoryMissingText +
+      ', Microsoft Building Shapefiles' + alwaysOptionalMissing;
+  const allMissingText = allMandatoryMissingText +
+      ', and must specify either damage asset or map bounds' +
+      allOptionalMissing;
+
+  it.only('validates asset data', () => {
```
juliexxia: rm only

docs/import/manage_disaster.js
janakdr: It's supposed to be different: gray vs yellow vs green. There was a bug where if it was yellow, when it was disabled it stayed yellow, so thanks for pointing that out! Added an assertion in the test for that case.

@ruthtalbot seemed ok with yellow as opposed to just green (that's my preference as well) in the chat.

docs/import/manage_disaster.js
janakdr: Prefer to not touch that file in this PR, that ok?

janakdr: Added a comment at the top.

docs/import/manage_disaster.js
janakdr: Agree it's getting hairy, wish I had an easier way. Can think about a refactoring soon.

docs/import/manage_disaster.js
juliexxia: Oh yeah, that sounds great too, as long as there's a difference between kicking off with all assets and kicking off with some optionals missing.

docs/import/manage_disaster.js
janakdr: https://github.com/givedirectly/Google-Partnership/pull/388

PR #319, opened 2019-12-16
Don't filter by damage bounds when counting buildings

janakdr:
The damage envelope can slice through a block group (especially if the block group is large), so just use the Join's implicit filtering on block groups to avoid dealing with buildings we don't want.

Tests were already pointing to this being an issue, should have thought more about it. Getting rid of the scaling in the tests seems to expose some EE inaccuracies, so had to loosen our floating-point checks to compensate.

### DIFF
--- a/docs/import/create_score_asset.js
+++ b/docs/import/create_score_asset.js
@@ -306,9 +306,8 @@ function createScoreAsset(
     }
 
     // Get building count by block group.
-    const buildingsHisto = buildingPath ?
-        computeBuildingsHisto(damageEnvelope, buildingPath, stateGroups) :
-        null;
+    const buildingsHisto =
+        buildingPath ? computeBuildingsHisto(buildingPath, stateGroups) : null;
 
     // Create final feature collection.
     const additionalTags = [];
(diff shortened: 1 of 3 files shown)


docs/import/create_score_asset.js
```diff
@@ -411,14 +410,12 @@ function calculateDamage(assetData, setMapBoundsInfo) {
...
- * @param {ee.Geometry.Polygon} damageEnvelope Area we are concerned with
  * @param {string} buildingPath location of buildings asset in EE
  * @param {ee.FeatureCollection} stateGroups Collection with block groups
  * @return {ee.Dictionary} Number of buildings per block group
  */
-function computeBuildingsHisto(damageEnvelope, buildingPath, stateGroups) {
-  const buildings =
-      ee.FeatureCollection(buildingPath).filterBounds(damageEnvelope);
+function computeBuildingsHisto(buildingPath, stateGroups) {
+  const buildings = ee.FeatureCollection(buildingPath);
```
juliexxia: So now it's all the buildings in the state? Have you run it/seen how it performs?

janakdr: It's not really, because we're doing the Join immediately below, which is basically what EE recommended we do for a geometry filter anyway. I haven't run it yet, but I don't see another correct option. Do you?

juliexxia: We could try a much bigger envelope than 1 km? Like try to get a sense of biggest block group in the US and go of that, even if it was way bigger than we needed still better than the whole state right. But good with this for now. Assuming this won't be so slow it kill EE right? We used to do this right?

janakdr:
Took 21 minutes to do all of Atlanta, so I think this is fine. The damage envelope is not intended to fully encompass all block groups, just to make it a little less "hole-y" when checking out neighborhoods.

I don't think the damage restriction was actually doing much, given the geo join immediately afterwards. The EE work might be largely the same in both cases.

PR #320, opened 2019-12-17
Back up previous version of score asset when creating new one

janakdr:
And use that backup if the new one is not found. We do this by catching a failure in our score promise and substituting the backup if the failure seems to be from a missing asset.

This works out because we don't do any processing of the score asset except via that promise. Stern comment added to that effect.

Test has a horrible wait because of an apparent Cypress bug (https://github.com/cypress-io/cypress/issues/5980).

### DIFF
--- a/cypress/integration/unit_tests/click_feature_test.js
+++ b/cypress/integration/unit_tests/click_feature_test.js
@@ -82,7 +82,7 @@ describe('Unit tests for click_feature.js with map and table', () => {
       containerDiv.appendChild(tableDiv);
       drawTableAndSetUpHandlers(
           convertEeObjectToPromise(scoredFeatures).then((fc) => fc.features),
-          map, features);
+          map);
     });
     cy.wrap(loadingFinishedPromise);
     return assertNoSelection();
--- a/cypress/integration/unit_tests/click_feature_test.js
+++ b/cypress/integration/unit_tests/click_feature_test.js
@@ -57,7 +57,7 @@ describe('Unit tests for click_feature.js with map and table', () => {
   beforeEach(() => {
     currentFeatures.clear();
     setUpPage();
-    cy.stub(Resources, 'getScoreAsset').returns(features);
+    cy.stub(Resources, 'getScoreAssetPath').returns(features);
     const oldConvertToPromise = MapUtil.convertEeObjectToPromise;
     MapUtil.convertEeObjectToPromise = (eeObject) => {
       MapUtil.convertEeObjectToPromise = oldConvertToPromise;
@@ -102,7 +102,6 @@ describe('Unit tests for click_feature.js with map and table', () => {
   }
 
   it('clicks on a feature on the map, then unclicks it', () => {
-    console.log('got here');
     clickAndVerifyBlockGroup();
     cy.wait(waitBeforeClick);
     cy.get('#test-map-div').click(500, 100);
--- a/cypress/integration/unit_tests/click_feature_test.js
+++ b/cypress/integration/unit_tests/click_feature_test.js
@@ -5,7 +5,10 @@ import * as loading from '../../../docs/loading.js';
 import {convertEeObjectToPromise} from '../../../docs/map_util.js';
 import {blockGroupTag, geoidTag} from '../../../docs/property_names';
 import {scoreTag} from '../../../docs/property_names.js';
-import {drawTableAndSetUpHandlers} from '../../../docs/run.js';
+import {
+  createScorePromise,
+  drawTableAndSetUpHandlers,
+} from '../../../docs/run.js';
 import {cyQueue} from '../../support/commands.js';
 import {loadScriptsBeforeForUnitTests} from '../../support/script_loader.js';
 import {convertPathToLatLng, createGoogleMap} from '../../support/test_map.js';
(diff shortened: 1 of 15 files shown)


janakdr:
This will probably require some extra work when #305 is implemented, because I think we'll want to track the score asset parameters for both the current asset and the backup. The reason for that is that we can be better about knowing things, like "is there damage for this asset?" if we keep asset metadata associated to both the current and previous ones. That way, if we fall back to previous, we'll have accurate data for the presence of damage.

docs/process_joined_data.js
```diff
@@ -1,6 +1,6 @@
 import {damageTag, scoreTag, snapPercentageTag} from './property_names.js';
 
-export {processJoinedData as default};
+export {processJoinedData};
```
juliexxia: why?

docs/resources.js
```diff
@@ -13,13 +14,30 @@ function getDisaster() {
...
+ * and inside run.js when constructing the score asset promise. Naively calling
+ * it could result in errors during the time when the score asset is being
+ * updated.
+ * @return {string}
+ */
 function getScoreAsset() {
-  // TODO(janakr): Modify map to handle no-damage scenario more intuitively.
-  //  Should make damage threshold, weight 0, probably hide those toggles
-  //  completely. Other changes?
+  // TOOD(janakr): rename this path.
```
juliexxia: Can you add to this comment what to rename to or incentive?

juliexxia: Maybe append Path to both getScoreAsset and getBackupScoreAsset?

docs/run.js
```diff
@@ -1,26 +1,75 @@
...
+
+/**
+ * Sets {@link snapAndDamagePromise} to the {@link ee.FeatureCollection} with
+ * id {@link getScoreAsset}, or, if that does not exist, with id
+ * {@link getBackupScoreAsset}.
+ *
+ * @return {Promise<string>} Promise with the name of the score asset found
+ */
+function setScorePromiseAndReturnAssetName() {
+  snapAndDamagePromise =
```
juliexxia: nit: maybe a good time to rename this var? to scorePromise or something?

docs/run.js
```diff
@@ -1,26 +1,75 @@
...
+}
+
+/**
+ * Calls {@link setScorePromiseAndReturnAssetName} and sets
+ * {@link resolvedScoreAsset} to the result. This is a separate function for use
+ * in tests.
+ */
+function setScorePromises() {
+  const scorePromiseAndReturnAssetName = setScorePromiseAndReturnAssetName();
+  resolvedScoreAsset.setPromise(scorePromiseAndReturnAssetName);
```
juliexxia: one line this?

docs/run.js
```diff
@@ -1,26 +1,75 @@
...
 const scalingFactor = 100;
 
+/**
+ * Contains `Promise<string>` with name of score asset. This is not just
+ * {@link getScoreAsset} because in the case that the asset with id
+ * {@link getScoreAsset} does not exist, we will fall back to the asset with id
+ * {@link getBackupScoreAsset}.
+ * @type {SettablePromise}
+ */
+const resolvedScoreAsset = new SettablePromise();
```
juliexxia: Does this need to be a settable promise? I thought that was for when it was possible you would want to "get" a promise" before "setting" it and I think that won't happen here? But maybe good for safety?

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -82,13 +95,14 @@ describe('Unit tests for click_feature.js with map and table', () => {
...
           convertEeObjectToPromise(scoredFeatures).then((fc) => fc.features),
-          map, features);
+          map);
     });
     cy.wrap(loadingFinishedPromise);
     return assertNoSelection();
   }
 
   it('clicks on a feature on the map, then unclicks it', () => {
+    console.log('got here');
```
juliexxia: debugging?

cypress/integration/unit_tests/run_test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+
+  it('Neither asset present', () => {
+    cy.stub(Resources, 'getScoreAsset')
+        .returns('nonexistent/feature/collection');
+    cy.stub(Resources, 'getBackupScoreAsset').returns('another/bad/asset');
+    const promise = setScorePromiseAndReturnAssetName().then(
+        (result) => {
+          throw new Error('unexpected: ' + result);
```
juliexxia: Do we want to throw an error here? Can we just assert result is null or something like that?

docs/run.js
janakdr: Done.

docs/run.js
janakdr: no longer needed.

docs/run.js
```diff
@@ -8,66 +8,55 @@ import {convertEeObjectToPromise} from './map_util.js';
...
-
-/**
- * Calls {@link setScorePromiseAndReturnAssetName} and sets
- * {@link resolvedScoreAsset} to the result. This is a separate function for use
- * in tests.
- */
-function setScorePromises() {
-  const scorePromiseAndReturnAssetName = setScorePromiseAndReturnAssetName();
-  resolvedScoreAsset.setPromise(scorePromiseAndReturnAssetName);
+  console.log(scorePromise);
```
juliexxia: debugging?

janakdr: Done!

Issue #277, opened 2019-12-05
Get rid of state asset pickers in manage_layers


janakdr:

juliexxia:
Noting this came up again as confusing in mock trial part 2

PR #327, opened 2019-12-17
Remove state asset pickers from manage layers page

ruthtalbot:


docs/import/manage_layers.js
```diff
@@ -328,31 +327,10 @@ function populateStateAndDisasterAssetPickers(disaster) {
   for (const state of states) {
```
janakdr: do we need this loop? And the code right above it?

juliexxia: I don't think so, I think the only promise that needs to be returned here is from disasterDone if that exists

docs/import/manage_layers.js
```diff
@@ -11,7 +11,6 @@ export {
   createAssetPickers,
   createLayerRow,
   createOptionFrom,
```
juliexxia: Not new, but createAssetPickers isn't actually imported anywhere (just checked on master) so could remove

docs/import/manage_layers.js
```diff
@@ -328,31 +327,10 @@ function populateStateAndDisasterAssetPickers(disaster) {
...
- * Create asset pickers for the given states.
- * @param {Array<string>} states of the form ['WA', ...]
- */
-function createStateAssetPickers(states) {
-  createAssetPickers(states, stateAssets, $('#state-asset-pickers'));
-}
-
 /**
  * Create an asset picker for the given disaster.
  * @param {string} disaster
```
juliexxia: could fold createDisasterAssetPicker and createAssetPickers into one function now since the former is the only caller of the latter (unless you think we might need it in the future).

docs/import/manage_layers.js
```diff
@@ -328,31 +327,10 @@ function populateStateAndDisasterAssetPickers(disaster) {
...
-    createStateAssetPickers(states);
-  } else {
-    const statesDone = getStatesAssetsFromEe(statesToFetch).then((assets) => {
-      for (const asset of assets) {
-        stateAssets.set(asset[0], asset[1]);
-      }
-      createStateAssetPickers(states);
-    });
-    promises.push(statesDone);
-  }
```
juliexxia: I also think you can get rid of the stateAssets map for this page entirely (but would double check that!)

docs/import/manage_layers.js
ruthtalbot: removed!

docs/import/manage_layers.js
ruthtalbot: Figured I'd leave them separate just in case, but happy to delete.

juliexxia: I'm actually doing the folding in my disable-empty-geometry PR so all good here

docs/import/manage_layers.js
```diff
@@ -321,13 +315,6 @@ function populateStateAndDisasterAssetPickers(disaster) {
     });
     promises.push(disasterDone);
```
juliexxia: Slightly simpler to return disasterDone here and either Promise.resolve() in the if block or allow this method to return nothing in the case that we didn't need to call getDisasterAssetsFromEe

docs/import/manage_layers.js
```diff
@@ -312,37 +306,15 @@ function populateStateAndDisasterAssetPickers(disaster) {
...
-      for (const asset of assets) {
-        stateAssets.set(asset[0], asset[1]);
-      }
-      createStateAssetPickers(states);
-    });
-    promises.push(statesDone);
-  }
-
-  return Promise.all(promises);
 }
```
juliexxia: Sorry if my other comments were confusing, I think you can delete createStateAssetPickers and its export!

janakdr:
Closes #277

PR #328, opened 2019-12-17
Don't delete backup asset unless we have to

janakdr:
Try a rename, if it fails because the backup was present, then delete, then try the rename again.

#306

### DIFF
--- a/docs/import/create_score_asset.js
+++ b/docs/import/create_score_asset.js
@@ -5,6 +5,9 @@ import {cdcGeoidKey, censusBlockGroupKey, censusGeoidKey, tigerGeoidKey} from '.
 
 export {createScoreAsset, setStatus};
 
+// For testing.
+export {backUpAssetAndStartTask};
+
 /**
  * For the given feature representing a block group, add properties for
  * total # buildings in the block group and # damaged buildings in the block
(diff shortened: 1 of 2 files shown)


docs/import/create_score_asset.js
```diff
@@ -319,47 +322,86 @@ function createScoreAsset(
...
-            // This check isn't perfect detection, but best we can do.
-            console.log(
-                'Old ' + scoreAssetPath + ' not found, did not move it');
-          } else {
-            const message = 'Error backing up: ' + err;
-            setStatus(message);
-            reject(new Error(message));
-            return;
-          }
+          console.error(message);
```
juliexxia: setStatus and console error?

docs/import/create_score_asset.js
```diff
@@ -319,47 +322,86 @@ function createScoreAsset(
...
       scoreAssetPath);
-  return new Promise((resolve, reject) => {
-    ee.data.deleteAsset(oldScoreAssetPath, (_, err) => {
-      if (err) {
-        if (err === 'Asset not found.') {
-          console.log(
-              'Old ' + oldScoreAssetPath + ' not present, did not delete it');
+  return renameAssetAsPromise(scoreAssetPath, oldScoreAssetPath)
+      .catch((renameErr) => {
+        console.log(renameErr);
```
juliexxia: always log err?

docs/import/create_score_asset.js
```diff
@@ -319,47 +322,86 @@ function createScoreAsset(
...
+              (deleteResolve, deleteReject) =>
+                  ee.data.deleteAsset(oldScoreAssetPath, (_, deleteErr) => {
+                    if (deleteErr) {
+                      // Don't try to recover here.
+                      // Don't show deletion error to user, but log it
+                      // somewhere.
+                      console.error(deleteErr);
+                      const message =
+                          'Error moving old score asset: ' + renameErr;
+                      setStatus(message);
```
juliexxia: Comment says don't show error but we're setting status here? Unclear how we'd ever reach this place right?

docs/import/create_score_asset.js
```diff
@@ -319,47 +322,86 @@ function createScoreAsset(
...
+/**
+ * Converts {@link ee.data.renameAsset} into a Promise, by resolving/rejecting
+ * in its callback.
+ * @param {string} from See {@link ee.data.renameAsset}
+ * @param {string} to See {@link ee.data.renameAsset}
+ * @return {Promise<?Object>} Promise that resolves if rename is successful,
+ *     and completes with rejection if there is an error. The rejection is
+ *     just a string, not an `Error` because that is what EarthEngine does. The
+ *     resolve will typically be with an Object that has information about the
+ *     renamed asset, but EarthEngine doesn't guarantee that
```
juliexxia: eurgh, any idea what it might be if not that object?

docs/import/create_score_asset.js
janakdr: debugging, removed.

docs/import/create_score_asset.js
janakdr: Clarified comment. We reach this place if there's an error deleting: there's a test case for it.

docs/import/create_score_asset.js
janakdr: debugging, removed.

docs/import/create_score_asset.js
janakdr: EE docs say it can be null. But doesn't matter, the returned value is irrelevant, because we know what the asset's name is going to be. We only use it for the error/non-error signal.

Issue #323, opened 2019-12-17
Check for non-null geometries on layer add


juliexxia:

janakdr:
Is there anything left here?

PR #329, opened 2019-12-17
Disable disaster folder FCs for those with non-null geometries and non FC state assets

juliexxia:
#323

This is mostly for manage_layers so people can't add geometry-less layers, but also affects the damage asset picker in manage_disaster. This is fine since we want the damage asset to have geometries anyway and it doesn't seem to slow it down too much. 

Still show filtered out assets but make them disabled. 

Initialize placeholder select on page load and then replace with one with the disabled assets when they come back from EE so the page doesn't jump around so much. 

Along the way, consolidate logic for asset disabling for both state and disaster drop downs inside their respect listEeAssets functions. If we want to do more disabling like, say, disabling image collections for the disaster asset (which we already do for states), we can refactor the methods to take a chainable then block or something. 

Also along the way, cache state directory calls to getStatesAssetsFromEe and convert manage_disaster's storage of state assets to store promises as well. 

Along the way fix a bug with sequential calls to getConvertEeObjectToPromiseRelease where we didn't know how to wait for preceding definitions of convertEeObjectToPromise to get put back before setting again. 

Should go in after #327. Would need to update some of the code deleted in that PR so (messily) disabled the relevant code for now to not do work we're about to throw away.


docs/import/list_ee_assets.js
```diff
@@ -35,7 +36,36 @@ function getDisasterAssetsFromEe(disaster) {
...
+                       const assetPromises = [];
+                       result.forEach((type, asset) => {
+                         if (type !== LayerType.FEATURE_COLLECTION) return;
+                         // assuming that if the first feature's geometry in a
+                         // collection is null, they're all null
+                         const assetPromise = convertEeObjectToPromise(
+                                             ee.FeatureCollection(asset)
+                                                 .first()
+                                                 .geometry()
+                                                 .coordinates())
```
janakdr: why coordinates and not just checking the geometry not null? That doesn't work?

docs/import/list_ee_assets.js
```diff
@@ -35,7 +36,36 @@ function getDisasterAssetsFromEe(disaster) {
...
+                                             .then((coords) => {
+                                               if (coords.length !== 0) {
+                                                 return [asset, type];
```
janakdr: nit: return null here instead of the implicit undefined? Prospective nit: in that case, just do a ternary, or do the positive `if (coords.length === 0) { return null; }`

docs/import/manage_layers.js
```diff
@@ -302,92 +303,70 @@ function createLayerRow(layer, index) {
...
  */
 function createStateAssetPickers(states) {
-  createAssetPickers(states, stateAssets, $('#state-asset-pickers'));
+  // Disabled for now waiting for #327
+  // createAssetPickers(states, stateAssets, $('#state-asset-pickers'));
 }
 
 /**
- * Create an asset picker for the given disaster.
+ * Set up the disaster asset picker div with a fake picker as a placeholder
```
janakdr: nit: Sets

docs/import/manage_layers.js
```diff
@@ -302,92 +303,70 @@ function createLayerRow(layer, index) {
...
+                               .append(assetPicker);
+  div.append(assetPickerLabel);
 }
+
 /**
- * Given either states or disasters, displays their assets in pickers.
- * @param {Array<string>} pickers e.g. ['WA',...] or ['harvey-2017']
- * @param {Map<string, Array<string>>} assetMap
- * @param {JQuery<HTMLElement>} div where to attach new pickers
+ * Display disaster assets in a select underneat the #disaster-adder-label label
```
janakdr: nit: Displays and "underneath"

docs/import/manage_layers.js
```diff
@@ -302,92 +303,70 @@ function createLayerRow(layer, index) {
...
+ * and add an add button which adds a layer.
```
janakdr: nit: and adds

janakdr:
If we go with the approach of waiting on all feature collections to be processed, I think it'll be more efficient if we do it server-side. That is, create an `ee.List` of all the geometries of the elements, then evaluate that, then look to see which elements are non-null. I'd also suggest doing an `ee.Algorithms.If(geometry, 1, null)` around each geometry, so that we don't have to fetch the actual geometries client-side, which could be pretty large.

juliexxia:
I can't tell if my one comment went through, but big updates in the PR description

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -100,6 +100,35 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
       ', and must specify either damage asset or map bounds' +
       allOptionalMissing;
 
+  it.only('has some disabled options', () => {
```
janakdr: Remove only

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -100,6 +100,35 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
...
+          ['asset2', {type: 2, disable: true}],
+        ])));
+    stateAssets.set('NY', new Map([
+      ['state0', {disable: false}],
+      ['state1', {disable: true}],
+    ]));
+    callEnableWhenReady(createDisasterData(['NY']));
+    cy.get('#select-asset-selection-row-poverty-NY > option')
+        .eq(2)
+        .invoke('attr', 'disabled')
```
janakdr: Can't just say should('be.disabled')?

docs/import/list_ee_assets.js
```diff
@@ -1,41 +1,117 @@
...
+/**
+ * Gets all assets in ee directories corresponding to given states. Caches
+ * results of calls to listEeAssets for each state. Marks assets with their type
+ * and whether or not they should be disabled when put into a select. Here,
+ * disabling any assets that aren't feature collections.
  * @param {Array<string>} states e.g. ['WA']
- * @return {Promise<Array<Array<string | Array<string>>>>} 2-d array of all
- *     retrieved assets in the form [['WA', {'asset/path': LayerType,...}], ...]
+ * @return {Promise<Array<Array<string, {disable: boolean}>>>} 2-d array of all
+ *     retrieved assets in the form [['WA', {'path': {disable: true}}],...]
```
janakdr: Are you anticipating more attributes to add to each asset? If not, why not just a boolean, instead of a dictionary?

juliexxia: Partial response to your partial review (: I thought since the disaster one has a full object because it needs the type as well the symmetry was nice. And I can see it being helpful if we ever want to expose the type. But can remove if you'd prefer.

docs/import/list_ee_assets.js
```diff
@@ -1,41 +1,117 @@
...
+            return Promise.all([
+              Promise.resolve(result),
+              convertEeObjectToPromise(ee.List(shouldDisable)),
+            ]);
+          })
+          .then((results) => {
+            const assetTypeMap = results[0];
+            const disableList = results[1];
+            const assetMap = new Map();
+            const assets = Array.from(assetTypeMap.keys());
```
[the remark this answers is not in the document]
janakdr: I was suggesting
```
const assetMapEntries = Array.from(assetTypeMap.entries());
for (let i = 0; i < assetMapEntries.length; i++) {
  const [asset, type] = assetMapEntries.get(i);
  ...
```

docs/import/list_ee_assets.js
janakdr: I dunno: we're silently filtering out folders, so it's not like we show everything blindly. feature collections without geometries are a harder case for the user to understand, knowing that you can't use an image as the SVI asset without it being explicitly disabled seems more like an expectation we could have. You can decide!

docs/import/list_ee_assets.js
```diff
@@ -1,41 +1,117 @@
...
   const promises = [];
   for (const state of states) {
-    const dir = legacyStateDir + '/' + state;
-    promises.push(listEeAssets(dir).then((result) => [state, result]));
+    let listEeAssetsPromise = stateAssetPromises.get(state);
+    if (!listEeAssetsPromise) {
+      listEeAssetsPromise = listEeAssets(legacyStateDir + '/' + state);
+      stateAssetPromises.set(state, listEeAssetsPromise);
+    }
+    promises.push(listEeAssetsPromise.then((result) => [state, result]));
```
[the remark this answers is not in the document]
janakdr: Can we also do this in listEeAssets? Especially since it's a library function.

docs/import/list_ee_assets.js
```diff
@@ -1,41 +1,117 @@
...
+                            .length(),
+                        ee.Number(0)),
+                    // DOM disable property can't recognize 0 and 1 as false and true :(
+                    true, false));
+              } else {
+                shouldDisable.push(false);
+              }
+            }
+            return Promise.all([
+              Promise.resolve(result),
```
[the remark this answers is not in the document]
janakdr: I'd rather assign it to an external variable

If you do keep it in the Promise.all, I think you don't need to wrap it. Just `Promise.all([result, ...` works.

docs/import/list_ee_assets.js
```diff
@@ -1,41 +1,111 @@
...
   const maybePromise = disasterAssetPromises.get(disaster);
   if (maybePromise) {
     return maybePromise;
   }
-  const result = listEeAssets(eeLegacyPathPrefix + disaster);
+  const result =
+      listEeAssets(eeLegacyPathPrefix + disaster)
+          .then((assetMap) => {
+            const shouldDisable = [];
+            for (const asset of Array.from(assetMap.keys())) {
```
janakdr: again, this could be `for (const [asset, type] of Array.from(assetMap.entries()))`

docs/import/manage_layers.js
```diff
@@ -302,92 +305,72 @@ function createLayerRow(layer, index) {
...
  */
 function createStateAssetPickers(states) {
-  createAssetPickers(states, stateAssets, $('#state-asset-pickers'));
+  // Disabled for now waiting for #327
+  // createAssetPickers(states, stateAssets, $('#state-asset-pickers'));
 }
 
 /**
- * Create an asset picker for the given disaster.
+ * Sets up the disaster asset picker div with a fake picker as a placeholder
```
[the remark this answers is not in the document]
janakdr: Sure, TODO sounds good.

cypress/integration/unit_tests/manage_layers_test.js
```diff
@@ -50,42 +54,99 @@ describe('Unit tests for manage_layers page', () => {
...
 
-  it('gets state asset info from ee', () => {
-    cy.wrap(getStatesAssetsFromEe([KNOWN_STATE])).then((assets) => {
-      // tests folder type asset doesn't make it through
-      expect(assets[0]).to.eql(
-          [KNOWN_STATE, new Map([[KNOWN_STATE_ASSET, 'TABLE']])]);
-      expect(ee.data.listAssets)
-          .to.be.calledWith(
-              legacyStatePrefix + KNOWN_STATE, {}, Cypress.sinon.match.func);
+  // TODO: fix this test.
```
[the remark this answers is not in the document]
juliexxia: Should be all fixed, also fixed the other test I wrote (in that previous PR) while this is in the front of my brain

cypress/integration/unit_tests/manage_layers_test.js
```diff
@@ -50,42 +54,99 @@ describe('Unit tests for manage_layers page', () => {
...
+  it.only('has racing disaster asset populates', () => {
```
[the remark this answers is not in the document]
juliexxia: I think this test covers that so hopefully all good!

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -393,20 +419,26 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
...
-    let secondRelease = getConvertEeObjectToPromiseRelease();
-    setFirstSelectInScoreRowTo(0, 'state1');
-
-    // release first evaluate but column still looks pending
-    firstRelease();
+    let firstRelease;
+    let firstStart;
+    cyQueue(() => {
+      const firstCall = getConvertEeObjectToPromiseRelease();
+      firstRelease = firstCall.releaseLatch;
```
janakdr: nit: use destructuring? Here and below

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -393,20 +419,26 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
...
+      firstStart = firstRelease.startPromise;
+    });
+    let secondRelease;
+    setFirstSelectInScoreRowTo(0, 'state0')
+        .then(() => firstStart)
+        .then(
+            () => secondRelease =
+                getConvertEeObjectToPromiseRelease().releaseLatch);
+    setFirstSelectInScoreRowTo(0, 'state1').then(() => firstRelease());
```
janakdr: nit: can just do `.then(firstRelease)`, I think. Here and below

cypress/integration/unit_tests/manage_layers_test.js
```diff
@@ -48,26 +51,99 @@ describe('Unit tests for manage_layers page', () => {
...
+    let finishedDisaster;
+    cy.document()
+        .then((doc) => {
+          const damageDiv = document.createElement('div');
+          damageDiv.id = 'disaster-asset-picker';
+          doc.body.appendChild(damageDiv);
+          cy.stub(document, 'getElementById')
+              .callsFake((id) => doc.getElementById(id));
+        })
+        .then(() => {
```
janakdr: nit: no need to have a separate .then() block here.

cypress/integration/unit_tests/manage_layers_test.js
```diff
@@ -48,26 +51,99 @@ describe('Unit tests for manage_layers page', () => {
...
+          firstConvertRelease = firstConvert.releaseLatch;
+          setCurrentDisaster(disaster);
+          finishedDisaster = getAssetsAndPopulateDisasterPicker(disaster);
+        });
+    let secondConvertRelease;
+    let finishedOther;
+    cy.get('#disaster-adder-label')
+        .find('select')
+        .should('be.disabled')
+        .then(() => {
```
janakdr: nit: can remove braces

cypress/support/import_test_util.js
```diff
@@ -85,19 +85,24 @@ function setUpSavingStubs() {
...
-      await promise;
-      return result;
-    });
+    startFunction();
+    return Promise
+        .all([MapUtil.convertEeObjectToPromise(eeObject), convertFinishLatch])
+        .then((results) => results[0]);
   };
-  return resolveFunction;
+  return {startPromise: convertStartPromise, releaseLatch: releaseLatch};
```
janakdr: nit: can omit `releaseLatch: `. And if you named "convertStartPromise" to "startPromise", could omit that as well.

docs/import/list_ee_assets.js
```diff
@@ -1,50 +1,119 @@
...
+
+/**
+ * Gets all assets in ee directories corresponding to given states. Caches
+ * results. Marks assets with their type and whether or not they should be
+ * disabled when put into a select. Here, disabling any assets that aren't
+ * feature collections.
+ * @param {Array<string>} states
+ * @return {Map<string, Promise<Map<string, {disabled: boolean}>>>}
  */
 function getStatesAssetsFromEe(states) {
```
janakdr: Doesn't need to happen in this PR, but at this point it's not clear why we request multiple states like this. Would make more sense to just request each state individually. The original reason, of listing all state folders and creating them if necessary, is long gone.

docs/import/list_ee_assets.js
```diff
@@ -53,14 +122,15 @@ function listEeAssets(dir) {
...
+ *     'users/gd/my-asset' -> 'IMAGE'
  */
 function getIds(listAssetsResult) {
-  const assets = new Map();
+  const assets = [];
   if (listAssetsResult.assets) {
     for (const asset of listAssetsResult.assets) {
       const type = maybeConvertAssetType(asset);
-      if (type) assets.set(asset.id, type);
+      if (type) assets.push({asset: asset.id, type: type});
```
janakdr: nit: can omit `type: `

docs/import/list_ee_assets.js
```diff
@@ -53,14 +122,15 @@ function listEeAssets(dir) {
 /**
  * Turns a listAssets call result into a map of asset -> type.
```
janakdr: nit: a bit stale?

docs/import/list_ee_assets.js
juliexxia: Are you imagining a promise per states instead?

janakdr: promise per individual state, yes. Aren't we basically already doing that here?

PR #338, opened 2019-12-18
Change stray client-side/static to docs/

janakdr:


README.md
```diff
@@ -69,7 +69,7 @@ https://dashboard.cypress.io in the "Runs" tab under "Settings".
 
     ```shell
     # Following line only needs to be run once per shell.
-    SOURCE_FILES='client-side/static/*.js 
+    SOURCE_FILES='web/*.js
```
ruthtalbot: Isn't this 'docs'? Feel like I'm missing something

janakdr: /facepalm

PR #344, opened 2019-12-18
Restrict access to manage_* pages to GD user

janakdr:
Also make some further cleanups to authenticate.js. Notably, get rid of annoying "Not logging into Firebase again" message.


docs/authenticate.js
```diff
@@ -195,32 +208,47 @@ function getFirebaseConfig(inProduction) {
...
 }
+
+/**
+ * Checks if a Firebase user is equal to the given Google user.
+ * @param {gapi.auth2.GoogleUser} googleUser
+ * @param {firebase.User} firebaseUser
+ * @return {boolean}
+ */
+function isUserEqual(googleUser, firebaseUser) {
+  if (firebaseUser) {
```
juliexxia: Do we know of a time when this can be fail?

docs/authenticate.js
```diff
@@ -50,35 +51,48 @@ function defaultErrorCallback(msg) {
...
-      initPromise.then(() => this.onLoginTaskCompleted());
+      this.gapiInitDone.setPromise(gapi.auth2.init(gapiSettings).then(() => {
+        const basicProfile =
+            gapi.auth2.getAuthInstance().currentUser.get().getBasicProfile();
+        if (this.needsGdUser &&
+            (!basicProfile || basicProfile.getEmail() !== gdUserEmail)) {
+          alert(
+              'You must be signed in as ' + gdUserEmail +
+              ' to access this page. Please open in an incognito window or ' +
+              'sign in as ' + gdUserEmail + ' here');
```
juliexxia: maybe s/here/in this window (can you hyperlink a here?)

janakdr: Rephrased. I don't think we need to hyperlink, since it will happen automatically when the alert closes. The follow-up PR will have that issue :/

docs/authenticate.js
janakdr: I think it might be null the first time you visit the page without having signed in before.

PR #349, opened 2019-12-20
Switch to using slightly more blessed method of getting EarthEngine tokens

janakdr:
Instead of logging into EE with our service account ourselves, we request an access token that can use EE (and nothing else). This was the recommended flow in a Google-internal support thread.

It was a real bear figuring out that we should use the googleapis library, and then how to use it. I'm not thrilled about using it, since it's not the new hotness (see https://www.npmjs.com/package/googleapis#support-and-maintenance), but it is actively maintained, so I think we're ok for a while at least.

As a drive-by non-code change, while experimenting with this, I significantly restricted our test-only service account's power. Rather than being an editor, it now only has permissions to create tokens and to manage Cloud Datastore (i.e., Firestore) data.

Ironically, we now no longer depend on the `@google/earthengine` node library. This may make IntelliJ a bit dumber about ee symbols, but it was never that smart to begin with.


cypress/plugins/main.js
```diff
@@ -104,21 +105,57 @@ function onFunction(on, config) {
           .then(() => firestoreUserToken)
           .finally(() => currentApp.delete());
     },
+
     /**
-     * Produces an EarthEngine token that can be used by production code.
+     * Produces an EarthEngine token that can be used by production code. We use
+     * the somewhat legacy and very poorly documented but still supported
```
juliexxia: Can we document anything we know about deprecation timeline and why we're not using the new stuff?

janakdr: Expanded comment a bit. No plans to deprecate, as I understand.

Issue #339, opened 2019-12-18
Make assets world-readable when adding them as layers


janakdr:

janakdr:
Is this done?

PR #354, opened 2019-12-20
Set EE asset to world readable when added as a layer

juliexxia:
#339


janakdr:
Might completely fix #339?

Issue #340, opened 2019-12-18
Allow anonymous/non-ee-whitelisted user to access page


janakdr:
I think that we can do this via the same method as in our tests. Stand up a simple server, that has service account credentials. The client-side code, if it failed to authenticate with EE, would request a token from that server, which would mint a new access token and return it. There are questions around how we would restrict who could get these tokens and how many they could get. But once the client-side code got a token, it could authenticate with EE similar to how our tests currently do it.

janakdr:
Only remaining issue here is actual deployment of the token server. Giving to Meylakh to advise on what he wants next steps to be. It's currently up and running on Google App Engine, but my understanding is that GD wants it on AWS.

janakdr:
Sounds like GD is ok with GAE as opposed to AWS.

PR #356, opened 2019-12-21
Handle Firestore authentication failures

janakdr:
Don't keep the map loading forever if Firestore user features fail to download. Instead, show a useful error message, and make the "User Features" line in the sidebar clickable, bringing up a dialog that will allow re-signing in, this time hopefully as a user with permissions.

Can be manually tested by going to the map as an EE-whitelisted but not Firestore-whitelisted user.

We can probably do a much better job around surfacing the error and what to do about it. Suggestions welcome, but probably best done in a follow-up.
Screen Shot 2019-12-21 at 4 50 52 PM
Screen Shot 2019-12-21 at 4 51 03 PM

#340


docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
  * @param {google.maps.Map} map Map to display regions on
  * @param {Promise<any>} firebasePromise Promise with Firebase damage data (also
  *     implies that authentication is complete)
- * @return {Promise} promise that is resolved when all initialization is done
- *     (only used by tests)
+ * @return {?google.maps.drawing.DrawingManager} DrawingManager added to map
```
juliexxia: Do you not document the implicit promise wrap for async functions?

docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
-        // gracefully.
-        damageAsset = doc.data()['asset_data']['damage_asset_path'];
-        userShapes = getFirestoreRoot().collection(collectionName);
-      })
-      .then(() => userShapes.get())
-      .then(
-          (querySnapshot) => drawRegionsFromFirestoreQuery(querySnapshot, map))
-      .catch(createError('getting user-drawn regions'));
+  try {
+    setUpPopup();
```
juliexxia: Does all of this need to be in this try block? Like can setUpPopup fail?

docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
+    let doc;
+    try {
+      doc = await firebasePromise;
+    } catch (err) {
+      // Firebase retrieval error handled elsewhere.
+      throw err;
```
juliexxia: Sorry if this is totally obvious, how is catching an error and immediately throwing it again different from just not catching the err?

docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
+    drawRegionsFromFirestoreQuery(querySnapshot, map);
+    return setUpPolygonDrawing(map);
+  } finally {
+    loadingElementFinished(mapContainerId);
+  }
+}
+
+const AUTHENTICATION_ERROR_CODE = 'permission-denied';
+const USER_FEATURES_DIALOG =
+    '<div>Sign in to authorized account to view user-drawn features</div>';
```
juliexxia: maybe \<p\> instead?

docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
+ * Handles errors thrown from retrieving user-defined features, most likely due
+ * to authentication issues.
+ * @param {Error|firebase.FirebaseError} err
+ */
+function handleUserShapesError(err) {
+  const userFeaturesRow = $('#' + userFeaturesCheckboxRowId);
+  userFeaturesRow.css('text-decoration', 'line-through');
+  // Checkbox just causes issues with double event triggering, and is useless
+  // anyway. Get rid of it.
+  userFeaturesRow.children('input').remove();
```
juliexxia: I think this looks a little weird. Possible to disable?

docs/polygon_draw.js
```diff
@@ -366,22 +387,79 @@ function setUpPolygonDrawing(map, firebasePromise) {
...
+        modal: true,
+        width: 600,
+        close: () => {
+          // Don't try to reuse the dialog.
+          dialogParent.empty();
+          dialogParent.remove();
+        },
+      });
+    };
+    userFeaturesRow.on('click', popUpDialog);
```
juliexxia: would it be possible to add a ? help icon like we're using elsewhere? Think it would go a long way to making it clear you can get help for whatever is happening here

docs/polygon_draw.js
janakdr: I don't think there's any harm in it. Basically, no matter what, we should make sure loadingElementFinished(mapContainerId) is called. Wrapping everything in the try means there's also no chance of programmer error, for instance if there's an early return.

docs/polygon_draw.js
janakdr: It isn't :) removed in favor of a comment.

docs/polygon_draw.js
janakdr: Is there a reason for \<p> over \<div?

docs/polygon_draw.js
janakdr: I wasn't able to get events on a disabled input to fire. https://stackoverflow.com/questions/3100319/event-on-a-disabled-input is similar to what I ran into, but it didn't work for me on Chrome either. That leads to the confusing situation where clicking on the checkbox does nothing, but clicking on the label brings up the dialog. Since I think users are likely to click on the checkbox if it's there, this seemed better.

Tried just making it not disabled. How's that?

docs/polygon_draw.js
janakdr: Nice idea, done.

janakdr: @ruthtalbot fyi gave you a TODO here

docs/polygon_draw.js
janakdr: Actually, seems like IntelliJ is fine either way, but then complains if you assign the result of the function to something which is supposed to be a Promise, so documented as a Promise.

docs/polygon_draw.js
janakdr: Can you say more about why that makes debugging easier? This try block doesn't catch anything, so there won't be any ambiguous messages about a thrown exception. I swapped setUpPopup() and addLoadingElement so the issue about unexpected exceptions leaving us vulnerable to loading forever should be moot.

docs/polygon_draw.js
janakdr: I think \<div> is better for a generic container. For instance, browsers add more space by default for \<p>
Screen Shot 2019-12-26 at 4 44 22 PM
versus \<div 
Screen Shot 2019-12-26 at 4 44 51 PM.
That's not actually what we want by swapping tags, and if we do want more space, setting it in the div is straightforward.

docs/polygon_draw.js
juliexxia: Ah so you're just doing it this way to make sure whatever gets returned we do the finish loading stuff? Got it.

PR #358, opened 2019-12-24
Handle errors loading layers

janakdr:
Show error to the user, disable the checkbox, and add mouseover text.

Add some basic EarthEngine error-message parsing and throw `Error` objects instead of strings, as Javascript prefers.


docs/layer_util.js
```diff
@@ -610,24 +621,47 @@ function removeScoreLayer() {
...
+            'Error loading layer ' + displayName);
+        $('#' + getCheckBoxRowId(index))
+            .prop(
+                'title',
+                'Error showing layer. If you believe the layer is there, try ' +
+                    'refreshing the page')
+            .css('text-decoration', 'line-through');
+        $('#' + getCheckBoxId(index))
+            .prop('checked', false)
+            .prop('disabled', true);
```
juliexxia: Woo great I'm seeing this after my comment in the other PR I think this is a lot better than just getting rid of the box entirely.

docs/layer_util.js
```diff
@@ -610,24 +621,47 @@ function removeScoreLayer() {
...
  * @return {Promise} wrappedPromise
  */
-function wrapPromiseLoadingAware(promise) {
+function wrapPromiseLoadingAware(promise, layerInfoForErrors) {
   addLoadingElement(mapContainerId);
-  return promise.then(mapLoadingFinished);
+  const {index, 'display-name': displayName} = layerInfoForErrors;
+  return promise
+      .catch((err) => {
+        showError(
```
juliexxia: I wrote this in the other PR, what do you think about adding a little help icon so it's more clear you can interact (will maybe try to click but hover will become clear) with a messed up layer

docs/layer_util.js
```diff
@@ -610,24 +621,47 @@ function removeScoreLayer() {
...
+            'Error with layer ' + displayName + ', ' + err,
+            'Error loading layer ' + displayName);
+        $('#' + getCheckBoxRowId(index))
+            .prop(
+                'title',
+                'Error showing layer. If you believe the layer is there, try ' +
```
juliexxia: can we persist some error message text in this hover text?

docs/checkbox_util.js
```diff
@@ -0,0 +1,20 @@
+export {getCheckBoxId, getCheckBoxRowId};
+
+/**
+ * Creates the id of a show/hide checkbox.
+ *
```
juliexxia: nit rm \/n ? I know we haven't been consistent with this

cypress/integration/unit_tests/checkbox_test.js
```diff
@@ -176,6 +178,103 @@ describe('Unit test for toggleLayerOn', () => {
...
+    };
+    addNullLayer(layer);
+    setUpLayerFailure().then(
+        () => toggleLayerOn(
+            {
+              'ee-name': 'asset/does/not/exist',
+              'asset-type': LayerType.FEATURE_COLLECTION,
+              'display-name': 'asset1',
+              'display-on-load': false,
+              'color-function': colorProperties,
```
juliexxia: why can't just user layer here? probably same answer, but can also reuse between tests?

cypress/integration/unit_tests/checkbox_test.js
```diff
@@ -176,6 +178,103 @@ describe('Unit test for toggleLayerOn', () => {
...
+      const div = doc.createElement('div');
+      div.id = getCheckBoxRowId(index);
+      const checkbox = doc.createElement('input');
+      checkbox.id = getCheckBoxId(index);
+      div.appendChild(checkbox);
+      doc.body.appendChild(div);
+      cy.stub(document, 'getElementById')
+          .callsFake((id) => doc.getElementById(id));
+    });
+    return cy.wrap(
```
juliexxia: difference between this strategy and using a stub constant that's accessible?

cypress/integration/unit_tests/checkbox_test.js
janakdr: Not sure I understand the question :) the asset type changes between two of the tests, didn't think worth extracting just for the remaining two tests for the second part.

cypress/integration/unit_tests/checkbox_test.js
janakdr: This doesn't pollute global namespace. Just a matter of preference. Slight advantage to this approach is you get Cypress execution order for "free", because of the `cy.get('@...'` on the other end.

docs/layer_util.js
janakdr: Don't think it's that useful here, because there's nothing you can do but reload the page. Especially now that your "globally accessible" PR is in. I guess we could add a section in the help to say to go to manage_layers.html and check that all your assets still exist? But until we have some idea what the user could do here, rather not add.

docs/layer_util.js
janakdr: This is different because nothing happens when you click on it, so disabling is easy.

docs/layer_util.js
janakdr: I think we've had a general tendency of pretty opaque errors in the user-visible part. I'm open to exposing more details, but would probably put it in the snackbar too if we're doing that. The error if the asset is not found looks like: `Collection asset 'users/gd/1999-janak/data-just-national-block-groups-19' not found.` If you think that would be useful, I can add it.

juliexxia: I think it'd be useful, maybe rephrased a tiny bit friendlier like "Underlying earth engine asset '...' not found." Just nice to have something that doesn't disappear (the snackbar fades eventually right?) / you can have multiple of at once.

cypress/integration/unit_tests/checkbox_test.js
juliexxia: Ah gotcha, missed some of the subtle differences. sgtm.

docs/layer_util.js
janakdr: What do you want to happen if the message doesn't end with 'not found.'? Suppress it, or show it verbatim?

juliexxia: I'd say show it verbatim since it's just hover text hopefully it won't be too annoying for anyone who can't understand it.

janakdr: Done. Pushed logic down into library function so it could be unified.

PR #359, opened 2019-12-24
Revamp update_test

janakdr:
update_test was using Promises but not actually waiting on them. This means it wasn't testing anything. In fact, there were three different failures visible in the console: 
update-errors

I redid it as follows:
* Used cy.document(), which allows you to actually see the form being created. Also enables us to interact in the typical Cypress way with the elements.
* Simplified the assertions about the arguments to createAndDisplayJoinedData.
* Stubbed out showError, so we could directly assert on it rather than trying to fake the page out.

I found a real bug while doing this: we were calling createAndDisplayJoinedData even if there was an error validating the arguments. I also noticed that we didn't clear the error div. We should probably just get rid of that div and rely on the snackbar, but leaving that alone for now.

There were also two errors in the tests, both asserting there were more inputs than there really were. Fixed those too.

### DIFF
--- a/cypress/integration/unit_tests/update_test.js
+++ b/cypress/integration/unit_tests/update_test.js
@@ -76,6 +76,10 @@ describe('Unit test for updates.js', () => {
             'have.text',
             'ERROR: poverty threshold must be between 0.00 and 1.00')
         .then(() => expect(errorStub).to.be.calledOnce);
+    cy.get('[id="poverty threshold"]').clear().type('0.0').blur();
+    cy.get('#update').click().then(
+        () => assertDisplayCalledWith(0.5, 0.0, 0.5));
+    cy.get('#error').should('have.text', '');
   });
 
   /**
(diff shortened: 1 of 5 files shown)


janakdr:
Needed to work around unpkg.com erroring out with @latest deck version. Hopefully will be fixed soon, but might as well pin for now, we're getting close to point where we should pin/download locally.

cypress/integration/unit_tests/update_test.js
```diff
@@ -1,137 +1,118 @@
...
-
-      expect(toggles.get('poverty weight')).to.equals(1);
-      expect($('#error').text()).to.equal('');
-      cy.wrap(createAndDisplayJoinedDataPromise).then(() => {
-        expect(lastPassedPovertyWeight).to.equals(1);
-        expect(lastPassedDamageThreshold).to.equals(0.0);
-      });
-    });
+    cy.wrap(setUpToggles(Promise.resolve({data: () => nullData}), {}));
+    cy.get('input').should('have.length', 2);
```
juliexxia: Ah is this from current settings button getting deleted

janakdr: Oh, that makes sense.

PR #361, opened 2019-12-24
Handle EE folder creation errors and Firestore errors in disaster creation

### LINKED ISSUE #331: Display error when EE folder fails to be created during manage_disaster add

janakdr:
We just show a snackbar error. Not sure there's more we can do, because it's unexpected.

This will slightly slow down creation, because it's EE then Firestore, but worth it.

Also fix slightly broken test: under the covers we were creating states for the letters in 'IG, MY' because inputs can't have array values.

Closes #331

### DIFF
--- a/cypress/integration/unit_tests/manage_disaster_add_delete_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_add_delete_test.js
@@ -3,11 +3,7 @@ import * as FirestoreDocument from '../../../docs/firestore_document.js';
 import {getFirestoreRoot} from '../../../docs/firestore_document.js';
 import {assetDataTemplate} from '../../../docs/import/create_disaster_lib.js';
 import {disasterData} from '../../../docs/import/manage_disaster';
-import {
-  addDisaster,
-  deleteDisaster,
-  writeNewDisaster,
-} from '../../../docs/import/manage_disaster.js';
+import {addDisaster, deleteDisaster, writeNewDisaster} from '../../../docs/import/manage_disaster.js';
 import {createOptionFrom} from '../../../docs/import/manage_layers.js';
 import {createAndAppend} from '../../support/import_test_util.js';
 import {loadScriptsBeforeForUnitTests} from '../../support/script_loader';
@@ -28,8 +24,11 @@ describe('Add/delete-related tests for manage_disaster.js', () => {
   let setAclsStub;
   beforeEach(() => {
     disasterData.clear();
-    createFolderStub = cy.stub(ee.data, 'createFolder').callsFake((dir, force, callback) => callback());
-    setAclsStub = cy.stub(ee.data, 'setAssetAcl').callsFake((id, acl, callback) => callback());
+    createFolderStub =
+        cy.stub(ee.data, 'createFolder')
+            .callsFake((asset, overwrite, callback) => callback());
+    setAclsStub = cy.stub(ee.data, 'setAssetAcl')
+                      .callsFake((asset, acls, callback) => callback());
   });
 
   it('writes a new disaster to firestore', () => {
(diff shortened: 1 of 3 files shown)


docs/import/manage_disaster.js
```diff
@@ -369,71 +370,67 @@ function addDisaster() {
...
-    folderCreationPromises.push(
-        getCreateFolderPromise(legacyStateDir + '/' + state));
+  const eeFolderPromises =
+      [getCreateFolderPromise(eeLegacyPathPrefix + disasterId)];
+  states.forEach(
+      (state) => eeFolderPromises.push(
+          getCreateFolderPromise(legacyStateDir + '/' + state)));
+
+  const tailError = '" You can try refreshing the page';
+  // Unsafe to do the Firestore write before EE folders are created, since maybe
```
juliexxia: super nit: I get this phrasing knowing the history of this method but maybe just say the affirmative i.e. "wait on ee folder creation to do firestore write because..."

docs/import/manage_disaster.js
```diff
@@ -369,71 +370,67 @@ function addDisaster() {
...
-    });
-    if (!added) disasterPicker.append(createOptionFrom(disasterId));
-    toggleState(true);
-
-    disasterPicker.val(disasterId).trigger('change');
+  try {
+    await disasterCollectionReference().doc(disasterId).set(currentData);
+  } catch (err) {
+    const message = err.message ? err.message : err;
+    showError('Error writing to Firestore: "' + message + tailError);
```
juliexxia: Should we do this here if we're not doing it everywhere?

docs/import/manage_disaster.js
```diff
@@ -369,71 +370,67 @@ function addDisaster() {
...
+    return false;
+  }
+  const disasterPicker = $('#disaster-dropdown');
+  const disasterOptions = disasterPicker.children();
+  let added = false;
+  // We expect this recently created disaster to go near the top of the list, so
+  // do a linear scan down.
+  // Note: let's hope this tool isn't being used in the year 10000.
+  // Comment needed to quiet eslint.
+  disasterOptions.each(/* @this HTMLElement */ function() {
```
juliexxia: nit: could inline disasterPicker.children() creation here and not create disasterOptions

docs/import/manage_disaster.js
```diff
@@ -369,71 +370,67 @@ function addDisaster() {
  * Writes the given details to a new disaster entry in firestore. Fails if
  * there is an existing disaster with the same details.
  *
```
juliexxia: maybe add comment detail about printing error since we removed from below?

cypress/integration/unit_tests/manage_disaster_add_delete_test.js
```diff
@@ -135,10 +150,64 @@ describe('Add/delete-related tests for manage_disaster.js', () => {
...
+    return $(document.createElement('option')).val(text).prop('selected', true);
+  }
+
+  it('Error creating EE folder', () => {
+    const id = '2005-summer';
+    const states = [KNOWN_STATE];
+    const errorStub = cy.stub(ErrorLib, 'showError');
+    // Tests don't have permission to create folders, so this will exercise EE
+    // failure mode better than we could fake it.
+    createFolderStub.restore();
```
juliexxia: why do you need to restore here?

janakdr: Because it was stubbed in the beforeEach. Expanded comment.

docs/import/manage_disaster.js
janakdr: Rephrased, how's this?

docs/import/manage_disaster.js
janakdr: perfect enemy of the good? This seems better than the prior state, and I don't think it will share code with the other write, if we do add error handling there.

docs/import/manage_disaster.js
janakdr: Not sure I understand. Saying here that this function now shows errors on failure? Added a couple of lines. Note that the prior comment, about not being the gd user, is no longer relevant since only the gd user can access this page.

cypress/integration/unit_tests/manage_disaster_add_delete_test.js
juliexxia: Ah sorry got this mixed up with resetHistory

PR #362, opened 2019-12-25
Set ee token server up to be deployed to Google App Engine

janakdr:
We add a `package.json` to `token_server/` to keep the server standalone, and give instructions to GAE about how to deploy it. That means that things that respect module boundaries no longer apply their settings to `token_server/`: something to be aware of.

`.gcloudignore` was auto-generated, didn't touch it.

#340


docs/authenticate.js
```diff
@@ -43,8 +43,9 @@ const eeErrorDialog =
     '<a href="https://signup.earthengine.google.com">https://signup.earthengine.google.com</a>' +
     ' or sign into a whitelisted account after closing this dialog</div>';
 
-// TODO(janakr): make configurable, especially for tests.
-const TOKEN_SERVER_URL = 'http://localhost:9080';
+const TOKEN_SERVER_URL = 'https://mapping-crisis.appspot.com';
```
juliexxia: cool!

token_server/README.md
```diff
@@ -1,4 +1,32 @@
 # Simple web server to provide EarthEngine tokens to non-whitelisted users
 
-Run with `node --experimental_modules token_server/token_server.js`.
+## Deploy to Google App Engine
 
+* Follow initial steps for Google App Engine deployment: install `gcloud`, log
+in as the GD user `gd-earthengine-user@givedirectly.org`, and set the gcloud
```
juliexxia: Can you specify where/what to log in to and where/how to set the gcloud project?

token_server/README.md
```diff
@@ -1,4 +1,32 @@
...
+project to `mapping-crisis`. Helpful resource:
+[NodeJS Quickstart](https://cloud.google.com/appengine/docs/standard/nodejs/quickstart).
```
juliexxia: Is there anything specific on this page that's relevant to the set up process you're documenting here? I think most/all of it has already been done and isn't pertinent? I agree it's helpful to link to some docs but would be awesome to link to some more specifically relevant places.

token_server/README.md
```diff
@@ -1,4 +1,32 @@
 # Simple web server to provide EarthEngine tokens to non-whitelisted users
 
-Run with `node --experimental_modules token_server/token_server.js`.
+## Deploy to Google App Engine
```
juliexxia: Can we add more of a header paragraph about who these instructions are for (maybe even date and person writing?) Mostly came to mind because I wouldn't want anyone who is earth engine whitelisted and just wants to access locally thinking they need this

janakdr: I don't think we expect anyone but developers to be viewing these READMEs, no? But added some text.

token_server/README.md
janakdr: I prefer not to duplicate widely available documentation, especially given Google's propensity to deprecate and replace. But added currently valid instructions and link to general SDK docs.

token_server/README.md
janakdr: I couldn't find anything more specifically relevant. But everything I did here came from that as a starting point. However, you're right, there's not much there. Removed.

token_server/README.md
juliexxia: Agreed only developers but developers like instructions too

token_server/README.md
juliexxia: It's not about being able to read documentation, more about finding the correct documentation and set up for our specific project. If there's a page that documents the steps you have here then we should just link that instead and hopefully that will get updated. I think it's nice to save users the steps of digging through different Google App Engine documentation pages if we can.

token_server/README.md
janakdr: I was addressing your "earth engine whitelisted" scenario: I don't think we expect anyone except Meylakh and Katie to be reading these docs. Alex Nawar shouldn't be getting here. Are these docs ok?

token_server/README.md
janakdr: I don't think there is anything more specific than what I linked to, at least not that I could find when looking again now. Would you like more detail than what is in the latest version of this file?

token_server/README.md
```diff
@@ -1,4 +1,37 @@
 # Simple web server to provide EarthEngine tokens to non-whitelisted users
 
-Run with `node --experimental_modules token_server/token_server.js`.
+This is an optional component for the mapping page, that allows any user to view
+it, even if they are not whitelisted with EarthEngine. Only the site maintainer
+should follow the instructions below, and they should only need to be done once.
 
+## Deploy to Google App Engine
+
+* Follow initial steps for Google App Engine deployment: install `gcloud`; when
```
juliexxia: maybe helpful to say 'install 'gcloud' (using homebrew or yarn)' or something to that effect to make it abundantly clear we mean the command line tool and the rest of these steps are for the terminal as well

token_server/README.md
```diff
@@ -1,4 +1,37 @@
...
+prompted, log into gcloud as the GD user `gd-earthengine-user@givedirectly.org`,
+and set the gcloud project to `mapping-crisis`. Typically you can run the
```
juliexxia: The bits after Typically are more specific repeats of the previous steps right? Might be nice to stick them in a legit code block

token_server/README.md
```diff
@@ -1,4 +1,37 @@
...
+commands `gcloud auth login` followed by
+`gcloud config set project mapping-crisis`. Please also consult the
+[Google Cloud SDK documentation](https://cloud.google.com/sdk/docs/).
+
+* Run `gcloud app deploy`. The app should now be running successfully at
+https://mapping-crisis.appspot.com (note that directly accessing that URL is not
+expected to work, since appropriate headers have to be attached to the request).
```
juliexxia: This means users would need to go to https://mapping-crisis.appspot.com/wherever/index.html is?

token_server/README.md
juliexxia: Added some comments!

token_server/README.md
juliexxia: I thought when you said you didn't want to duplicate documentation there was documentation saying the same steps as you'd written here, in which case we should just link to it but if there's not then nice to leave the steps IMO

token_server/README.md
janakdr: Linked to gcloud installation instructions.

token_server/README.md
janakdr: Rewrote. I realized the reason there wasn't specific documentation was that the commands themselves are basically self-documenting: if you run a command without running a previous one, it tells you what you need to run. So got rid of this and just gave the last command to run. Now I'm not worried about documentation getting out of date.

token_server/README.md
janakdr: This server is intended to be a separate component that isn't a hard dependency of the general mapping page. I added a paragraph in the README intro. Does it make more sense now?

PR #363, opened 2019-12-27
Get rid of looking states up in batches

janakdr:
We don't need it anymore under the covers, so simplifies code.

Also get rid of stray `.only`.

### DIFF
--- a/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
+++ b/cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
@@ -2,7 +2,7 @@ import {addPolygonWithPath} from '../../../docs/basic_map.js';
 import {readDisasterDocument} from '../../../docs/firestore_document.js';
 import {createDisasterData} from '../../../docs/import/create_disaster_lib.js';
 import * as ListEeAssets from '../../../docs/import/list_ee_assets.js';
-import {assetSelectionRowPrefix, disasterAssets, disasterData, scoreAssetTypes, scoreBoundsMap, setUpScoreBoundsMap, setUpScoreSelectorTable, stateAssets, validateUserFields} from '../../../docs/import/manage_disaster';
+import {assetSelectionRowPrefix, disasterAssets, disasterData, scoreAssetTypes, scoreBoundsMap, setUpScoreBoundsMap, setUpScoreSelectorTable, validateUserFields} from '../../../docs/import/manage_disaster';
 import {enableWhenFirestoreReady} from '../../../docs/import/manage_disaster.js';
 import {getDisaster} from '../../../docs/resources.js';
 import {cyQueue} from '../../support/commands.js';
@@ -25,6 +25,7 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
   loadScriptsBeforeForUnitTests('ee', 'firebase', 'jquery', 'maps');
   before(preparePage);
 
+  let stateStub;
   setUpSavingStubs();
   let firstTest = true;
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -39,8 +40,8 @@ describe('Score parameters-related tests for manage_disaster.js', () => {
         firstTest = false;
       }
     });
+    stateStub = cy.stub(ListEeAssets, 'getStateAssetsFromEe');
```
juliexxia: use as notation for this stub? might as well start moving towards that?

docs/import/list_ee_assets.js
```diff
@@ -2,44 +2,38 @@ import {eeLegacyPathPrefix, legacyStateDir} from '../ee_paths.js';
...
 import {convertEeObjectToPromise} from '../map_util.js';
 
-export {getDisasterAssetsFromEe, getStatesAssetsFromEe};
+export {getDisasterAssetsFromEe, getStateAssetsFromEe};
 
 /**
- * Cache for the results of getStatesAssetsFromEe for each state.
- * @type {Map<string, Promise<Map<string, string>>>}
+ * Cache for the results of getStateAssetsFromEe.
+ * @type {Map<string, Promise<Map<string, {disabled: boolean}>>>}
```
juliexxia: thanks!

docs/import/manage_disaster.js
```diff
@@ -255,30 +246,16 @@ function onSetDisaster() {
...
-  }
-  const scorePromise = Promise.all(statePromises).then((stateAssets) => {
-    if (getDisaster() === currentDisaster &&
-        !processedCurrentDisasterStateAssets) {
-      // Don't do anything unless this is still the right disaster.
-      initializeScoreSelectors(states, stateAssets);
-      processedCurrentDisasterStateAssets = true;
-    }
-  });
+  // getStateAssetsFromEe does internal caching.
```
juliexxia: does this mean we don't need disasterAssets either because that also does internal cacheing? If so can we leave a TODO?

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
janakdr: I'm not all in on `as` notation for stubs: I think it makes sense where the stub is being passed "around the back" between two functions, but nothing else needs to know about it. Also makes sense where the stub is being asserted on after test functions execute, so you want to enforce Cypress order. Neither is true for this stub, so prefer to leave it as is.

That being said, if you would like to have only one way that we do things, so generally get rid of these "let" instance variables and move everything to `as`, including for non-stubs, I'm open to it.

docs/import/manage_disaster.js
janakdr: Great point, TODO added.

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
juliexxia: I think it'd be nice to unify but it's low priority for me so fine to leave it this way

docs/import/manage_disaster.js
juliexxia: Might be nice to leave similar TODO in manage_layers but understand if you don't want to touch that file in this PR. Maybe could leave a general clean up callers TODO in list_ee_assets. Though I already have a TODO so unify manage_layers and manage_disasters so

janakdr: I promise to do them together :)

janakdr: #397

Issue #336, opened 2019-12-18
Make color picker start with column then discrete/continuous


janakdr:

janakdr:
Anything left here, or can we close this?

PR #365, opened 2019-12-27
Add a 'color uniformly' or 'color by property' flow to color function setting

juliexxia:
We now record the last 'by-property' style chosen in firebase. This is only used for when we toggle back and forth between coloring uniformly and coloring by property so the inner state of coloring by property is preserved. 

Do a little more creation in the HTML since it's less straightforward now across schemas. 

Along the way, fix a bug where if we chose discrete coloring but didn't actually pick any colors, the color td was empty (should be a transparent box). 

Warn when missing fields needed to properly display on map. Annotate these fields with a red asterisk. In the case of discrete, warn if any colors are missing.

Do some updating to not use single-quote var names where we can now that we're in camelCase and use more jquery-chai in tests.


docs/import/color_function_util.js
```diff
@@ -21,10 +20,18 @@ let globalTd;
...
-  const colorFunctionDiv = $('#color-fxn-editor');
-  colorFunctionDiv.prepend(createRadioFor(ColorStyle.SINGLE));
-  colorFunctionDiv.prepend(createRadioFor(ColorStyle.DISCRETE));
-  colorFunctionDiv.prepend(createRadioFor(ColorStyle.CONTINUOUS));
+  colorStyleTypeStrings.forEach(
+      (typeAsString, colorStyle) =>
+          $('#' + typeAsString + '-radio')
+              .on('change', () => switchSchema(colorStyle)));
+  $('#property-radio').on('change', () => {
+    console.log('changing');
```
janakdr: debugging

docs/import/manage_layers.html
```diff
@@ -56,10 +56,19 @@
       </table>
     </div>
     <div id="color-fxn-editor" class="editor content-container" hidden="true">
-      <div id="color-radios"></div><br>
+      <div>
+        <input type="radio" id="SINGLE-radio" name="property-or-single"><label for="SINGLE-radio" class="radio-label">color uniformly</label>
+        <input type="radio" id="property-radio" name="property-or-single"><label for="property-radio" class="radio-label">color by property</label>
```
janakdr: I thought we discussed swapping these? That is, first selecting a property, then selecting continuous/discrete, since that seems more intuitive for a user (they're more likely to know what column they want than to say, "I really want this dataset to indicate something continuous").

If going that way, there's a nice avenue for future work to make a reasonable guess about continuous/discrete based on the column properties. Can at least have a TODO in this PR.

docs/import/color_function_util.js
```diff
@@ -344,6 +336,9 @@ function switchSchema(type) {
 function displaySchema(type) {
   const colorFunction = getColorFunction();
   $('.color-type-div').hide();
+  const byPropertyDiv = $('#by-property');
+  const continuousDiv = $('#continuous');
+  const discreteDiv = $('#discrete');
   switch (type) {
     case ColorStyle.SINGLE:
       $('#single-color-picker').val(colorFunction['color']);
```
janakdr: In the case that no color has been chosen yet, this has weird behavior. The select shows no text, and the td is clear: 
Screen Shot 2019-12-27 at 12 05 21 PM
But clicking on the select shows that red is selected: 
Screen Shot 2019-12-27 at 12 05 33 PM
But if you don't actually click on anything, it remains clear.

I don't think this is new in this PR, but since you're dealing with unset values, what do you think of generating a default color, perhaps based on the layer name so it's deterministic but not always the same, at layer creation? That way there's a sensible default, as opposed to having this confusing "non-color" state that's only reachable right after layer creation. Could also add a TODO if you think it's a reasonable idea but want to leave for later.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -10,30 +13,130 @@ let writeToFirebaseStub;
...
+          wings: {min: 0, max: 100, values: [0, 1, 2, 100]},
+        },
+        field: 'wings',
+        colors: {},
+      },
+    });
+    td.trigger('click');
+    td.trigger('click');
+    expect(snackbarStub.withArgs(
+               'Warning: Closed layer missing at least one color. ' +
```
janakdr: oh, did we want this, or did we want it to warn if *all* colors were missing? @ruthtalbot

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -42,8 +145,8 @@ describe('Unit tests for color function utility', () => {
     const maxMin = $('#max-min');
     expect(maxMin.is(':visible')).to.be.false;
 
-    const continuousPropertyPicker = $('#continuous-property-picker');
-    continuousPropertyPicker.val('wings').trigger('change');
+    const propertyPicker = $('#property-picker');
```
janakdr: nit: could inline this, right? Could also use Cypress .get() (here and elsewhere, since creating inside Cypress doc).

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -189,7 +367,7 @@ function expectOneFirebaseWrite() {
 }
 
 /**
- * Gets the current color function.
+ * Gets the current color function.Fscore-
```
janakdr: stray characters?

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -79,23 +182,83 @@ describe('Unit tests for color function utility', () => {
...
-        'colors': {},
-        'color': 'yellow',
+        colors: {},
+        color: 'yellow',
       },
     };
     const td = setUpWithLayer(layer);
-    expect(colorFunctionEditor.is(':visible')).to.be.false;
-
+    expect($(colorFunctionEditor).is(':visible')).to.be.false;
```
janakdr: I think Cypress chai has jQuery assertions, so even if you're not using Cypress selectors, you can get more meaningful assertions: https://docs.cypress.io/guides/references/assertions.html#Chai-jQuery

docs/import/add_layer.js
```diff
@@ -51,21 +44,34 @@ function processNewEeLayer(asset, type) {
...
   }
 }
 
+/**
+ * Create fields that all ee layers have.
+ * @param {string} asset
+ * @param {LayerType} type
+ * @return {Object}
+ */
+function createCommonColorFunctionFields(asset, type) {
```
janakdr: nit: this function name talks about color, but isn't really about color. Maybe "createCommonLayerFields"?

docs/import/color_function_util.js
```diff
@@ -21,62 +21,70 @@ let globalTd;
...
+    setProperty(property);
+    switch (getColorFunction().currentStyle) {
+      case ColorStyle.CONTINUOUS:
+        maybeDisplayMinMax(property);
+        break;
+      case ColorStyle.DISCRETE:
+        populateDiscreteColorPickers(property);
+        break;
+      case ColorStyle.SINGLE:
+        throw Error('Somehow tried to set property while in single color mode');
```
janakdr: nit: will this get caught anywhere? Consider showing a snackbar error instead/in addition.

docs/import/color_function_util.js
```diff
@@ -266,49 +253,96 @@ function withColor(td, layer, property) {
...
+  // open td other than most recent closed
+  if (colorFunctionDiv.is(':visible')) {
+    maybeDisplayWarningOnClose();
+  }
+  colorFunctionDiv.show();
   selectCurrentRow(false);
   globalTd = td;
   selectCurrentRow(true);
+  const {currentStyle} = getColorFunction();
   $('#' + colorStyleTypeStrings.get(type) + '-radio').prop('checked', true);
```
janakdr: looks like type variable still needed here. I'm getting JavaScript errors loading page now.

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: Inlined. I would rather stick with current syntax as to not have to chain non-cypressing things to get them in the right order if that's alright with you.

docs/import/color_function_util.js
juliexxia: Sorry, forgot to rename that one. Thanks!

docs/import/color_function_util.js
juliexxia: Now with more snacks!

docs/import/color_function_util.js
```diff
@@ -21,62 +21,73 @@ let globalTd;
...
   const continuousColorPicker = createColorPicker('continuous-color-picker');
   continuousColorPicker.on('change', () => setColor(continuousColorPicker));
-  const continuousPropertyPicker =
-      $(document.createElement('select'))
-          .prop('id', 'continuous-property-picker');
-  continuousPropertyPicker.on('change', () => {
-    setProperty(continuousPropertyPicker);
-    maybeDisplayMinMax(continuousPropertyPicker);
+
+  // TODO: make an educated guess about if this property should
```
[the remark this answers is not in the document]
juliexxia: Added a TODO

docs/import/color_function_util.js
```diff
@@ -395,18 +442,18 @@ const discreteColorPickerDataKey = 'value';
...
-  if (!colorFunction['field']) {
+  const field = colorFunction.field;
+  if (!field) {
     return;
   }
-  const values =
-      colorFunction['columns'][$('#discrete-property-picker').val()]['values'];
+  const values = colorFunction['columns'][field]['values'];
   if (values.length === 0) {
     $('#too-many-values').show();
```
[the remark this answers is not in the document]
juliexxia: As discussed offline, this is because I wasn't hiding it property inside populateDiscreteColorPickers. Should be fixed now.

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -12,44 +14,150 @@ describe('Unit tests for color function utility', () => {
...
+            'Warning: layer missing all colors. May not show up on map.');
+    // discrete with everything
+    $('#discrete-color-pickers')
+        .find('select')
+        .first()
+        .val('red')
+        .trigger('change');
+    expect(missingFieldsWarning).not.to.be.visible;
+
+    // user doesn't have a way to set to pickers so need to make a whole new
```
janakdr: nit: sorry, I can't parse this

cypress/integration/unit_tests/color_function_util_test.js
```diff
@@ -107,35 +275,40 @@ describe('Unit tests for color function utility', () => {
...
 
     // switch to continuous
-    $('#CONTINUOUS-radio').trigger('change');
+    const propertyRadio = $('#property-radio');
+    propertyRadio.prop('checked', true).trigger('change');
     expectOneFirebaseWrite();
-    const continuousPropertyPicker = $('#continuous-property-picker');
+
+    const continuousRadio = $('#CONTINUOUS-radio');
+    expect(continuousRadio).to.have.prop('checked', true);
```
janakdr: nit: this is fine, but I think you can just say `expect(continuousRadio).not.to.be.checked`. Here and below

docs/import/color_function_util.js
```diff
@@ -266,49 +296,59 @@ function withColor(td, layer, property) {
...
+  colorFunctionDiv.show();
   selectCurrentRow(false);
   globalTd = td;
   selectCurrentRow(true);
-  $('#' + colorStyleTypeStrings.get(type) + '-radio').prop('checked', true);
-  displaySchema(type);
+  const {currentStyle} = getColorFunction();
+  $('#' + colorStyleTypeStrings.get(currentStyle) + '-radio')
+      .prop('checked', true);
+  if (currentStyle !== ColorStyle.SINGLE) {
```
janakdr: why not unconditionally trigger changed for the radio above? Won't that do what you want here?

juliexxia: not exactly, trigger the change for the continuous radio, for example, will do everything but check the by property radio. This is only an issue on color editor open since on regular schema change, you can't switch to continuous or discrete without the property radio already being checked.

cypress/integration/unit_tests/color_function_util_test.js
juliexxia: My bad, forgot a word

docs/import/color_function_util.js
```diff
@@ -16,33 +16,35 @@ export {
...
 let globalTd;
 
-const getWarning = (missing) =>
-    'Warning: layer missing ' + missing + '. May not show up on map.';
-
 /**
  * Displays a warning in color function editor if field or color(s) are missing
- * @param {string|number|boolean} field
- * @param {string|boolean} color
+ * @param {string|number|boolean} hasField
```
janakdr: nit: why are these named like booleans but can be strings? Maybe just use !! in the callsites?

docs/import/color_function_util.js
janakdr: Ah, gotcha.

janakdr:
#336

PR #367, opened 2019-12-27
Try to make score assets world readable every time GD user loads any page

janakdr:
This should help even if they don't remember to make readable, although if the first visit to the map is by another user after the score asset is created, we're still vulnerable.

I tried to use the *IamPolicy functions rather than the deprecated *AssetAcl functions, but they're terrible enough that might need to go back. Will wait a few days to hear from EE support list.

Did drive-by cleanup of `SettablePromise` and `TaskAccumulator` imports, and noticed that `script.js` was being way too baroque. `SettablePromise` might be on its last legs.


docs/authenticate.js
```diff
@@ -276,6 +288,65 @@ function initializeFirebase() {
...
+ *
+ * Optimized for the common case of assets already being world-readable: the
+ * getIamPolicy is technically not necessary, but usually it will avoid a set,
+ * which should be at least as expensive.
+ */
+function makeScoreAssetsWorldReadable() {
+  // TODO(janakr): Switch to using listEeAssets once #368 is in.
+  // TODO(janakr): Consider sharing cache with list_ee_assets.js. Not trivial
+  //  because that code does additional EE requests to look at geometries, so
+  //  we would need a two-level cache, one raw and one with geometries.
```
juliexxia: I had the idea that you could pass in an extra filtering function to listEeAssets that does something like geometry filtering and would really generify listEeAssets. This is potentially compelling for that direction as well.

docs/authenticate.js
```diff
@@ -276,6 +288,65 @@ function initializeFirebase() {
...
+        }
+        const paths = new Set([getScoreAssetPath(), getBackupScoreAssetPath()]);
+        let foundAssets = 0;
+        for (const {id} of listResult.assets) {
+          if (paths.has(id)) {
+            foundAssets++;
+            ee.data.getIamPolicy(eeLegacyPrefix + id, () => {})
+                .then((policy) => {
+                  for (const binding of policy.bindings) {
+                    if (binding.role === 'roles/viewer') {
```
juliexxia: maybe this is documented somewhere, what is this check for?

docs/authenticate.js
```diff
@@ -276,6 +288,65 @@ function initializeFirebase() {
...
+                        ee.data.setIamPolicy(
+                            eeLegacyPrefix + id, policy, () => {});
+                      }
+                      return;
+                    }
+                  }
+                  // TODO(janakr): Do better. See what EE says.
+                  // If we got here, no roles/viewer binding. Use some
+                  // Javascript magic.
+                  const BindingConstructor = policy.bindings[0].constructor;
```
juliexxia: Agreed this feels like it can't be best practice.

docs/authenticate.js
```diff
@@ -276,6 +288,65 @@ function initializeFirebase() {
...
+                  policy.bindings.push(new BindingConstructor(allReadBinding));
+                  ee.data.setIamPolicy(eeLegacyPrefix + id, policy, () => {});
+                });
+          }
+          if (foundAssets === paths.size) {
```
juliexxia: factor out paths.size so we don't need to get it every time?

cypress/integration/unit_tests/authenticate_test.js
```diff
@@ -0,0 +1,214 @@
...
+  // Give other thread a chance to call listAssets if that's going to happen.
+  cy.wait(0).then(() => expect(getPolicyStub).to.not.be.called);
+});
+
+/**
+ * Asserts that all the library calls we expect to encounter have happened. For
+ * conciseness, we use the `this` mode of accessing Cypress-aliased variables
+ * (https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Sharing-Context).
+ * This means that this function cannot be called from within an arrow function.
+ * @this Context
```
juliexxia: Any prefixes you can add to this Context to describe more where it's coming from

cypress/integration/unit_tests/authenticate_test.js
```diff
@@ -0,0 +1,214 @@
...
+  };
+}
+
+it('Does not try to make score assets readable when not gd user', () => {
+  cy.get('@gapiEmail')
+      .then((emailStub) => emailStub.returns('some-user@givedirectly.org'));
+  const getPolicyStub = cy.stub(ee.data, 'getIamPolicy');
+  waitForTrackAndAssertNormalStubs();
+  // Give other thread a chance to call listAssets if that's going to happen.
+  cy.wait(0).then(() => expect(getPolicyStub).to.not.be.called);
```
juliexxia: cyQueue here? Is cy.wait(0) and cy.wrap(null) slightly different?

docs/authenticate.js
janakdr: Sure, we could do that, but there's still the question of how to cache. You'd have to cache keyed on filtering function. I think the simplest thing to do is probably just have two caches, one of the raw listings and one of the processed listings, but still not positive it's worth it.

docs/authenticate.js
janakdr: Sorry, most of this was from trial and error/poking around in objects that I got back in the console. But added links in jsdoc above.

docs/authenticate.js
janakdr: Thanks, done.

cypress/integration/unit_tests/authenticate_test.js
janakdr: Yes, they're slightly different. See https://github.com/givedirectly/Google-Partnership/blob/master/cypress/integration/unit_tests/cyqueue_test.js#L18 for a hint of the behavior. `cy.wrap()` and `cy.get()` do not explicitly give up control to another thread: if the object inside the wrap is an ordinary one, or if the element is findable on the page, respectively, they continue execution in the current thread. That means that if another thread is waiting to go, it won't get a chance. `cy.wait()` will always give up control: `cy.wait(0)` puts itself right back on the event loop stack, but will allow anything else that is waiting to run to go first.

That's why, in the test I linked to, even though the Promise had resolved before both of them, the `.then` clause for `cy.get()` ran before the Promise's `.then` clause, while the one for `cy.wait()` ran after.

cypress/integration/unit_tests/authenticate_test.js
janakdr: Done, sorry, got lazy.

PR #368, opened 2019-12-27
Fix latent bug in ee_list_assets when too many assets present

janakdr:


docs/import/ee_utils.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+function listAssetsRecursive(path, result, pageToken) {
+  return ee.data
+      .listAssets(path, pageToken ? {page_token: pageToken} : {}, () => {})
+      .then((listAssetsResult) => {
+        if (!listAssetsResult) {
+          return;
+        }
+        if (listAssetsResult.assets) {
+          result.push(...listAssetsResult.assets);
```
juliexxia: Can you bulk up this doc description a little more. I originally thought this was recursive in the sense that it looks inside folders for more assets or something like that

docs/import/list_ee_assets.js
```diff
@@ -2,7 +2,9 @@ import {eeLegacyPathPrefix, legacyStateDir} from '../ee_paths.js';
 import {convertEeObjectToPromise} from '../ee_promise_cache.js';
 import {LayerType} from '../firebase_layers.js';
 
-export {getDisasterAssetsFromEe, getStateAssetsFromEe};
+import {listEeAssets} from './ee_utils.js';
+
```
juliexxia: nit: rm extra \n

docs/import/ee_utils.js
janakdr: Changed to be iterative and use some Javascript sugar. Abusing your approval to submit, but please let me know if there are issues and I can address in follow-up!

Issue #335, opened 2019-12-18
Name score asset better, and make it easier to create layers from it(?)


janakdr:

janakdr:
For instance, have a separate button just for the score asset, above the other disaster assets.

PR #369, opened 2019-12-28
Rename score asset to poverty-damage-score

janakdr:
Don't rename backup: if anything, we should make backup less attractive, so people don't try to create assets from it. Leaving that for another day. For now, it's alphabetically not that close to poverty-damage-score, so that's something.

I've already copied data-ms-as-tot over to poverty-damage-score for all of our disasters.

#335

### DIFF
--- a/docs/resources.js
+++ b/docs/resources.js
@@ -28,7 +28,7 @@ function getDisaster() {
 function getScoreAssetPath() {
   // TODO(janakr): rename this path to something more intelligible, like
   //  .../score-asset
-  return gdEePathPrefix + getDisaster() + '/data-ms-as-tot';
+  return gdEePathPrefix + getDisaster() + '/poverty-damage-score';
 }
 
 /**


janakdr:
I copied all the existing assets to the new name. Is that what you're asking about? Layers will keep on using the old name. I guess I could change them?

janakdr:
Since this submission is mostly independent of changing layers, I'll do this submission now and change the layers in a bit.

janakdr:
For posterity, did this in `manage_layers_lib#setDisasterData`:

```
  for (const data of readDisasterData.values()) {
    for (const layer of data.layers) {
      if (layer['ee-name']) {
        layer['ee-name'] = layer['ee-name'].replace(/\/data-ms-as-tot$/, '/poverty-damage-score');
      }
    }
  }
```

Then opening each disaster and triggering a save by focusing an element overwrote the asset path.

Issue #322, opened 2019-12-17
Make damage percentage undefined ('-'?) instead of 0 for no-damage score asset


juliexxia:

janakdr:
@ruthtalbot I think all that's left here is whether we want to surface that the damage asset is not present at all. Currently it'll just be omitted completely, no longer "damage percentage: undefined". Can also ask the trickier question for fields that are present for some features but not others: should we show an indication. Repurposing this issue for that, and giving to you for a decision. If we do want to show that a feature is missing fields compared to other features, let me know and will implement, otherwise will mark as fixed.

ruthtalbot:
Is the second question the same thing we discussed a while ago of whether to hide fields in the popup if data is missing for that feature? If I'm understanding that, my preference here would be to show text that says "data unavailable" or something equivalent so the user knows why they are seeing that data in one pop up but not the other.

For damage, because we allow a score to be calculated without it, I think leaving it out completely is fine. If we do want to surface it, I could see putting a message under the toggles sidebar that says "add damage data to see damage in map" or something equivalent, but it feels excessive to put a "damage unavailable" message in every single popup.

Issue #17, opened 2019-08-02
Consider flexible data sources


janakdr:
A number of feature requests from GiveDirectly pertain to bringing in specific data sources. Since the best data sources may change over time, it might be helpful to rephrase this request as the ability to import arbitrary (already well-formatted) data sources and display them as desired. As a starting point, we could provide an interface where arbitrary geotiff (raster) data could be uploaded by GD admins and viewable as a layer. Same for arbitrary Shapefile data.

It gets more complicated once that data has to be used to compute a score, rather than as an overlay. But I think we might be able to do something interesting there as well with a preprocessing pipeline that extracts the necessary properties from such data. In any case, the above feature already seems like a good one to have.

janakdr:
The layer aspect of this is done. The score part is definitely still an issue, and something I'm thinking about now.

janakdr:
This is basically done. The only remaining issue is that score asset creation seems to be slower than state-based, especially with buildings, to a level that might make it unworkable. I don't really understand why it's slower, though. Will continue to investigate if I have time.

PR #370, opened 2019-12-28
Get properties of features dynamically, don't hard-code them

### LINKED ISSUE #304: Make it so score asset isn't dependent on SVI/Income

janakdr:
Start to make map much more flexible. Instead of a hard-coded list of properties, since we're looping over the features anyway, keep track of their properties and display the ones we found. This will allow us to create a very generic score asset, with new properties we may never have considered, and display them on the map. In the limit, GD should be able to run score asset creation with a very general poverty asset.

Don't display absent properties in pop-up ("absent" now means, "absent from this feature, but present in other features"). I can't remember if we decided this was the preferred way to go, but I think it was. If we want to display '-', can modify to do that as well.

Make subtle change so that if damage tag is absent, will still do calculation properly. Can take advantage of that in a follow-up in create_score_asset.js to not set the damage tag when damage not present.

Big downside of this PR is that property names are sorted alphabetically (except for some manually curated ones). I think the trade-off is worth it. If we really want to preserve ordering of currently known columns, we could do that (keep known columns sorted, if they're present, and add unknown ones at the end).

Closes #304
#322, #17

### DIFF
--- a/cypress/integration/unit_tests/click_feature_test.js
+++ b/cypress/integration/unit_tests/click_feature_test.js
@@ -2,10 +2,7 @@ import {tableContainerId} from '../../../docs/dom_constants.js';
 import {convertEeObjectToPromise} from '../../../docs/ee_promise_cache.js';
 import {currentFeatures} from '../../../docs/highlight_features';
 import * as loading from '../../../docs/loading.js';
-import {
-  blockGroupTag,
-  geoidTag
-} from '../../../docs/property_names';
+import {blockGroupTag, geoidTag} from '../../../docs/property_names';
 import {scoreTag} from '../../../docs/property_names.js';
 import * as Resources from '../../../docs/resources.js';
 import {drawTableAndSetUpHandlers, resolveScoreAsset} from '../../../docs/run.js';
@@ -85,8 +82,17 @@ describe('Unit tests for click_feature.js with map and table', () => {
       tableDiv.id = 'table';
       containerDiv.appendChild(tableDiv);
       drawTableAndSetUpHandlers(
-          convertEeObjectToPromise(scoredFeatures).then((fc) => (
-              {featuresList: fc.features, columnsFound: [geoidTag, blockGroupTag, scoreTag, 'OTHER PERCENTAGE', 'SOME PROPERTY']})),
+          convertEeObjectToPromise(scoredFeatures).then((fc) => ({
+                                                          featuresList:
+                                                              fc.features,
+                                                          columnsFound: [
+                                                            geoidTag,
+                                                            blockGroupTag,
+                                                            scoreTag,
+                                                            'OTHER PERCENTAGE',
+                                                            'SOME PROPERTY',
+                                                          ],
+                                                        })),
           map);
     });
     cy.wrap(loadingFinishedPromise);
--- a/cypress/integration/unit_tests/click_feature_test.js
+++ b/cypress/integration/unit_tests/click_feature_test.js
@@ -179,7 +179,7 @@ describe('Unit tests for click_feature.js with map and table', () => {
     cy.get('#test-map-div').should('contain', 'SCORE: 1');
     cy.get('#test-map-div').should('contain', 'my block group');
     // Rounded because property name ends with 'PERCENTAGE'.
-    cy.get('#test-map-div').should('contain', 'OTHER PERCENTAGE: 4.232');
+    cy.get('#test-map-div').should('contain', 'OTHER PERCENTAGE: 4.233');
     cy.get('#test-map-div').should('contain', 'SOME PROPERTY: 100');
     cy.get('.google-visualization-table-tr-sel')
         .find('[class="google-visualization-table-td"]')
@@ -252,6 +252,6 @@ function createFeatureWithOnlyGeoid(west, south, east, north) {
 function createFeatureFromCorners(west, south, east, north) {
   let result = createFeatureWithOnlyGeoid(west, south, east, north);
   result = result.set('SOME PROPERTY', 100);
-  result = result.set('OTHER PERCENTAGE', 4.23223434);
+  result = result.set('OTHER PERCENTAGE', 4.23293434);
   return result;
 }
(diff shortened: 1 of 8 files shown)


janakdr:
@ruthtalbot let me know if you have any thoughts about the high-level direction of this PR.

janakdr:
The table headings are listed alphabetically, and the items in the popup are alphabetical top to bottom, except for the first three, which are hard-coded. See 
image
image

Re missing items, there are two aspects. One is whether the overall data set has all "expected" properties, like SVI and income. To do that, we intrinsically have to hard-code expected properties. I'm not excited about that idea, because it seems to forestall flexibility. The other is whether an individual district is missing data that other districts have. I'm happy to surface that, as I said in the description, but I think it will also be somewhat rare. (The only way we know of for it to happen is if you have SVI, for instance, for one state and not another, and in a non-per-state world, that seems even less likely.) If that's what we're talking about in terms of surfacing missing data, I can make the change in this PR.

docs/import/create_score_asset.js
```diff
@@ -1,13 +1,19 @@
...
 export {createScoreAsset, setStatus};
-
 // For testing.
 export {backUpAssetAndStartTask};
 
+const TRACT_TAG = 'TRACT';
+const SVI_TAG = 'SVI';
+// Median household income in the past 12 months (in 2016 inflation-adjusted
+// dollars)
+const INCOME_TAG = 'MEDIAN INCOME';
```
juliexxia: Is it always 2016 inflation-adjusted dollars or is it just the year you download from

docs/import/create_score_asset.js
```diff
@@ -109,9 +115,9 @@ function combineWithSnap(feature, snapKey, totalKey) {
         snapFeature.get(censusGeoidKey),
         blockGroupTag,
         snapFeature.get(censusBlockGroupKey),
-        snapPopTag,
+        povertyHouseholdsTag,
```
juliexxia: We're not actually displaying snap households in the table/popup right now but it makes sense to add it I guess.

docs/polygon_draw.js
```diff
@@ -109,10 +109,11 @@ class StoredShapeData {
     const numDamagePoints = StoredShapeData.prepareDamageCalculation(polygon);
     const intersectingBlockGroups =
         StoredShapeData.getIntersectingBlockGroups(polygon);
+    // TODO(janakr): we should have poor/total
```
juliexxia: What does this TODO mean?

docs/process_joined_data.js
```diff
@@ -22,8 +22,9 @@ const scoreDisplayCap = 255;
  */
 function colorAndRate(
     feature, scalingFactor, povertyThreshold, damageThreshold, povertyWeight) {
-  const povertyRatio = feature.properties[snapPercentageTag];
-  const ratioBuildingsDamaged = feature.properties[damageTag];
+  const povertyRatio = feature.properties[povertyPercentageTag];
+  // If damage tag is absent, treat as 0.
+  const ratioBuildingsDamaged = feature.properties[damageTag] || 0;
```
juliexxia: Was this already happening but just more safe now?

docs/process_joined_data.js
```diff
@@ -41,6 +42,11 @@ function colorAndRate(
   feature.properties['color'] = [255, 0, 255, opacity];
 }
 
+const KEY_BLACKLIST = Object.freeze(new Set(['system:index', 'color']));
```
juliexxia: Can you add comments about why we expect these two strings to be here and why we don't want to show them. Maybe constant-ize 'color' and use above?

docs/process_joined_data.js
```diff
@@ -60,17 +66,28 @@ function colorAndRate(
...
+                povertyWeight,
+              },
+            ]) => {
+        const columnsFound = new Set(ALWAYS_PRESENT_KEYS);
         for (const feature of featuresList) {
           colorAndRate(
               feature, scalingFactor, povertyThreshold, damageThreshold,
               povertyWeight);
+          for (const key of Object.keys(feature.properties)) {
+            if (!columnsFound.has(key) && !KEY_BLACKLIST.has(key)) {
```
juliexxia: does the set not already handle the !columnsFound.has(key) part of this logic?

docs/process_joined_data.js
```diff
@@ -60,17 +66,28 @@ function colorAndRate(
 function processJoinedData(
```
juliexxia: Can you beef up the return type here

cypress/integration/unit_tests/click_feature_test.js
```diff
@@ -84,7 +82,17 @@ describe('Unit tests for click_feature.js with map and table', () => {
...
+                                                              fc.features,
+                                                          columnsFound: [
+                                                            geoidTag,
+                                                            blockGroupTag,
+                                                            scoreTag,
+                                                            'OTHER PERCENTAGE',
+                                                            'SOME PROPERTY',
+                                                          ],
+                                                        })),
           map);
```
juliexxia: test for these headers to show up somewhere?

janakdr: Good idea, done.

docs/import/create_score_asset.js
janakdr: Just moved this from property_names.js. Removed.

docs/import/create_score_asset.js
janakdr: Easier than hiding, I agree.

docs/polygon_draw.js
janakdr: Rephrased.

docs/process_joined_data.js
janakdr: Nope, it would have been `undefined` and then the calculations below would have been bogus.

docs/process_joined_data.js
janakdr: Reworked logic a bit, made more general.

docs/process_joined_data.js
janakdr: Done.

docs/process_joined_data.js
janakdr: Done. The "not being able to use add instead of contains" throws me off.

PR #371, opened 2019-12-28
Add support to token server for AWS Elastic Beanstalk

janakdr:
Main code change is that, because we're no longer running as a user that we can just give Google permissions to (the Google App Engine user), we have to store and load the credentials ourselves. The server can't do anything until that happens, so we wrap everything in the promise for that.

Other changes:
* AWS only supports Node 10 (boo) so had to drop down. Maybe they'll offer Node 12 soon: they offered 10 on [11/19/18](https://aws.amazon.com/about-aws/whats-new/2018/11/aws-elastic-beanstalk-adds-support-for-nodejs-10/), 20 days after Node 10 became the ["active" version](https://nodejs.org/en/about/releases/). Node 12 became the active version 10/21/19.
* Realized we were unnecessarily re-initializing auth and client objects inside ee_token_creator.js. Keep them around forever now.
* Was double-initializing currentTokenPromise, both directly and via generateTokenPeriodically. No longer.

#340

### DIFF
--- a/token_server/README.md
+++ b/token_server/README.md
@@ -34,7 +34,7 @@ unless it is getting traffic.
 [Dashboard](https://console.cloud.google.com/appengine?folder=&organizationId=838088520005&project=mapping-crisis).
 You must be logged in as `gd-earthengine-user@givedirectly.org` to access it.
 
-### Set up Google Cloud instance
+### Set up new Google Cloud instance
 * Most of the necessary work needs to be done to set up the mapping page to work
 properly in the first place. The only additional work is to create a service
 account that is whitelisted for EarthEngine access (currently
@@ -60,7 +60,7 @@ run `eb deploy` after running `eb init` in order to log in.
 are using Elastic Beanstalk.
 
 ### Set up new Amazon Elastic Beanstalk instance
-* See the [Google Cloud instructions](#set-up-google-cloud-instance) if that is
+* See the [Google Cloud instructions](#set-up-new-google-cloud-instance) if that is
 also changing.
 * Delete the `.elasticbeanstalk` subdirectory of this directory, and follow the
 instructions [here]((https://aws.amazon.com/getting-started/tutorials/deploy-app-command-line-elastic-beanstalk/)
--- a/token_server/README.md
+++ b/token_server/README.md
@@ -34,9 +34,44 @@ unless it is getting traffic.
 [Dashboard](https://console.cloud.google.com/appengine?folder=&organizationId=838088520005&project=mapping-crisis).
 You must be logged in as `gd-earthengine-user@givedirectly.org` to access it.
 
-## Run server on Amazon EC2
+### Set up Google Cloud instance
+* Most of the work needs to be done to set up the mapping page to work properly
+in the first place. The only additional work is to create a service account that
 
-`¯\_(ツ)_/¯`
+## Run server on Amazon Elastic Beanstalk
+
+* Deploy using [these instructions](https://aws.amazon.com/getting-started/tutorials/deploy-app-command-line-elastic-beanstalk/).
+Since the project already exists on Amazon Elastic Beanstalk, you just need to
+run `eb deploy` after running `eb init` in order to log in. TODO(janakr): check this
+
+### Set up Amazon Elastic Beanstalk instance
+* See the [Google Cloud instructions](#set-up-google-cloud-instance) if that is
+also changing.
+* Delete the `.elasticbeanstalk` subdirectory of this directory, and follow the
+instructions [here]((https://aws.amazon.com/getting-started/tutorials/deploy-app-command-line-elastic-beanstalk/)
+to set up from scratch.
+* Delete the `.gitignore` file that setup creates, and add `.elasticbeanstalk/*`
+to git.
+* Add Google service account credentials to
+[Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/).
+Remember that you need to be logged into Google as
+`gd-earthengine-user@givedirectly.org`.
+   * The Google service account is currently
+`earthengine-token-provider@mapping-crisis.iam.gserviceaccount.com` (specified
+in `ee_token_creator.js`).
+   * Download private key JSON for the Google service account via the
+   [Credentials page](https://console.developers.google.com/apis/credentials?project=mapping-crisis).
+   * Upload to the [Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/)
+   using the provvided instructions.
+To make the secret readable by the server, you will have to have the server run
+as a user with permissions to read the secret:
+  * [Instructions for creating the user](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/iam-instanceprofile.html#iam-instanceprofile-create).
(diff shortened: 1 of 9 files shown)


token_server/README.md
```diff
@@ -34,9 +34,59 @@ unless it is getting traffic.
 [Dashboard](https://console.cloud.google.com/appengine?folder=&organizationId=838088520005&project=mapping-crisis).
 You must be logged in as `gd-earthengine-user@givedirectly.org` to access it.
 
-## Run server on Amazon EC2
+### Set up new Google Cloud instance
+* Most of the necessary work needs to be done to set up the mapping page to work
```
juliexxia: Having slight trouble parsing this first sentence

token_server/README.md
```diff
@@ -34,9 +34,59 @@ unless it is getting traffic.
...
+* Deploy using [these instructions](https://aws.amazon.com/getting-started/tutorials/deploy-app-command-line-elastic-beanstalk/).
+Since the project already exists on Amazon Elastic Beanstalk, you just need to
+run `eb deploy` after running `eb init` in order to log in.
+
+* The server URL will be entered here once we have GiveDirectly's account.
+
+* See the [console](https://console.aws.amazon.com/) for more information. We
+are using Elastic Beanstalk.
+
+### Set up new Amazon Elastic Beanstalk instance
```
juliexxia: Can you add a when/why you would set up a new instance?

token_server/README.md
```diff
@@ -34,9 +34,59 @@ unless it is getting traffic.
...
+to set up from scratch.
+* Delete the `.gitignore` file that setup creates, and add `.elasticbeanstalk/*`
+to git.
+* Add Google service account credentials to
+[Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/).
+Remember that you need to be logged into Google as
+`gd-earthengine-user@givedirectly.org`.
+   * The Google service account is currently
+`earthengine-token-provider@mapping-crisis.iam.gserviceaccount.com` (specified
+in [ee_token_creator.js](./ee_token_creator.js).
```
juliexxia: missing a closing paren here

token_server/README.md
```diff
@@ -34,9 +34,59 @@ unless it is getting traffic.
...
+   * Download private key JSON for the Google service account via the
+   [Credentials page](https://console.developers.google.com/apis/credentials?project=mapping-crisis).
+   * Upload to the [Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/)
+   using the provided instructions.
+To make the secret readable by the server, you will have to have the server run
+as a user with permissions to read the secret:
```
juliexxia: is the colon supposed to indicate the next line of instructions? Maybe double indent the next two lines?

token_server/app.yaml
```diff
@@ -1 +1 @@
-runtime: nodejs12
+runtime: nodejs10
```
juliexxia: Maybe add a todo to update to 12 if/when amazon supports?

token_server/aws_get_credentials.js
```diff
@@ -0,0 +1,46 @@
+import AWS from 'aws-sdk';
+import fs from 'fs';
+import tmp from 'tmp';
+
+export {storeGoogleCredentials};
+
+const AWS_REGION = 'us-east-1';
+const SECRET_ID = 'arn:aws:secretsmanager:us-east-1:560508828482:secret:' +
```
juliexxia: document or links to docs?

token_server/README.md
janakdr: Rephrased, pointing to future INITIAL_SETUP doc. better?

token_server/README.md
janakdr: Well, Meylakh is about to set up a new instance. It's not routine: the main use case would be if a new user was setting up or they were changing their Amazon account for some reason. I expect that if someone comes to this section, it's because they have to do it, for whatever reason. Added a line.

token_server/README.md
janakdr: Yeah, I wasn't doing proper md formatting. Should be better now.

token_server/app.yaml
janakdr: Done.

token_server/aws_get_credentials.js
janakdr: Done, and realized I had left that out of docs, so important catch!

janakdr:
The whole token server setup does not remove/break anything if the user is EE-whitelisted: we won't try to contact it. This PR allows the server to run on AWS. Before this PR, it could only run on Google Cloud (Google App Engine, specifically). I think Julie's question might have been why we needed to support AWS as well.

token_server/README.md
```diff
@@ -24,69 +24,81 @@ you a command to run, probably `gcloud auth login`. Run it, logging in as
...
+   Remember that you need to be logged into Google as
+   `gd-earthengine-user@givedirectly.org`.
+     -  The Google service account is currently
+        `earthengine-token-provider@mapping-crisis.iam.gserviceaccount.com`
+        (specified in [ee_token_creator.js](./ee_token_creator.js)).
+     -  Download private key JSON for the Google service account via the
+        [Credentials page](https://console.developers.google.com/apis/credentials?project=mapping-crisis).
+     -  Upload to the [Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/)
+        using the provided instructions.
+     -  Take the "Secret ARN"'" (starting with `arn:aws:secretsmanager`) and put
```
juliexxia: something going on with the quotes here?

token_server/README.md
```diff
@@ -24,19 +24,81 @@ you a command to run, probably `gcloud auth login`. Run it, logging in as
...
+        it into the `SECRET_ID` field in
+        [aws_get_credentials.js](./aws_get_credentials.js).
+     - Make sure the region you created the secret in matches the region
+       specified in `AWS_REGION` in
+       [aws_get_credentials.js](./aws_get_credentials.js)! 
+   To make the secret readable by the server, you will have to have the server
```
juliexxia: This paragraph is coming up as part of the previous bullet (at least according to the in review md preview page)

token_server/README.md
```diff
@@ -24,19 +24,81 @@ you a command to run, probably `gcloud auth login`. Run it, logging in as
...
+   Remember that you need to be logged into Google as
+   `gd-earthengine-user@givedirectly.org`.
+     -  The Google service account is currently
+        `earthengine-token-provider@mapping-crisis.iam.gserviceaccount.com`
+        (specified in [ee_token_creator.js](./ee_token_creator.js)).
+     -  Download private key JSON for the Google service account via the
+        [Credentials page](https://console.developers.google.com/apis/credentials?project=mapping-crisis).
+     -  Upload to the [Amazon Secrets Manager](https://aws.amazon.com/secrets-manager/getting-started/)
+        using the provided instructions.
+     -  Take the "Secret ARN"'" (starting with `arn:aws:secretsmanager`) and put
```
juliexxia: something going on with "/' s here?

token_server/README.md
janakdr: Fixed, let me know if flow looks ok to you.

PR #372, opened 2019-12-30
Document initial app setup

### LINKED ISSUE #167: Document/track non-source changes made to set up system
This is unofficial documentation since I don't know the right place for it.

Random things:

In https://console.cloud.google.com/apis/credentials?organizationId=838088520005&project=mapping-crisis, I modified the "Browser key" to only be usable by the Maps JavaScript API, Places API, Google Earth Engine API, and Identity Toolkit API. This seems to work for me, but I may have missed an API that can be called. 
screenshot-api

There should probably be a separate issue for creating a public-facing API key that is restricted to be used on only givedirectly.github.io, but not quite sure how that will work or what the risks are.

I've set up github to serve our site from the master branch, docs directory: 
github pages settings

janakdr:
Add an INITIAL_SETUP.md doc that explains how the system is set up behind the scenes. It's in Markdown because setting up the system requires making source code changes anyway, so there's no additional barrier imposed by keeping the docs in GitHub.

Shove our Firestore rules into source control, which is a big relief for me. We can now properly track and recover them.

Mild source comments and refactorings to make documentation a bit clearer.

Markdown probably best reviewed by just going to https://github.com/givedirectly/Google-Partnership/blob/document-setup/INITIAL_SETUP.md and reading it through.

Closes #167


INITIAL_SETUP.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Moving to a different server for the mapping page
+
+Our current server setup is GitHub Pages. We serve from the `master` branch,
+`docs` folder:
+![github pages setup screenshot](./markdown_images/github_pages.png)
+
+Because the application is all client-side JavaScript, any static web host
+(WordPress, a blog, etc.) should work as a server. Simply copy all the files
+from [docs/](./docs) and below to the location you're serving from.
```
ruthtalbot: What does the 'and below' mean here? Just any files mentioned below this text?

INITIAL_SETUP.md
```diff
@@ -0,0 +1,203 @@
+# Instructions for Initial Setup
+
+This is a reference for the steps taken to set this app up the first time. None
+of these steps should have to be performed again, but they may be a useful
```
ruthtalbot: Little neurotic, but can we bold "None of these steps..." 

Just want to be very sure no one takes this as required and starts repeating these.

INITIAL_SETUP.md
```diff
@@ -0,0 +1,203 @@
...
+    -  Give it a useful name (the current one is "Disaster relief mapping").
+  
+    -  Specify "Authorized JavaScript origins". Currently they are
+       `http://localhost:8080` for local development and
+       `https://givedirectly.github.io`. This second origin should be the domain
+       that the mapping page is hosted on.
+  
+    -  Specify "Authorized redirect URIs". You must list here every page that a
+       normal user or the Admin user might go to in the app, including when
+       managing a disaster. Currently, it is `http://localhost:8080`,
```
ruthtalbot: Do we need to have the image collection page here if it's to be usable? Or would they just do that locally?

INITIAL_SETUP.md
```diff
@@ -0,0 +1,203 @@
...
+-  Commit, push, and merge this change, including the new file
+   `service_account.json.enc`.
+
+-  [Enable EarthEngine API](https://console.cloud.google.com/apis/api/earthengine.googleapis.com/overview)
+   for Test account in the developer console.
+
+### Modify/set up Firestore rules
+
+-  One-time: install `firebase-tools` globally:
+   `yarn global add firebase-tools`.
```
ruthtalbot: Do they need to install yarn?

INITIAL_SETUP.md
janakdr: Done.

INITIAL_SETUP.md
janakdr: Rephrased, better?

INITIAL_SETUP.md
janakdr: yes, if it's usable. Although the image collection page is kind of its own beast in terms of the authenticator (since it needs to use GCS) so we'll have to see how it gets made usable.

INITIAL_SETUP.md
janakdr: Yes, that's in the main README instructions too, but I added it here too.

INITIAL_SETUP.md
```diff
@@ -0,0 +1,204 @@
+# Instructions for Initial Setup
+
```
juliexxia: Might be nice to date the last edit of the page for the non-github savvy

janakdr: I agree it would be nice, but since the reader has to modify files in github anyway, and forgetting to update the date seems likely, will leave out for now.

PR #374, opened 2019-12-31
Use listEeAssets instead of raw ee.data.listAssets

janakdr:


docs/authenticate.js
```diff
@@ -312,47 +313,42 @@ const allReadBinding = Object.freeze({
...
-                  // If we got here, no roles/viewer binding. Use some
-                  // Javascript magic.
-                  const BindingConstructor = policy.bindings[0].constructor;
-                  policy.bindings.push(new BindingConstructor(allReadBinding));
-                  ee.data.setIamPolicy(eeLegacyPrefix + id, policy, () => {});
-                });
-          }
-          if (foundAssets === numAssets) {
-            return;
+  listEeAssets(eeLegacyPathPrefix + getDisaster()).then((listedAssets) => {
```
juliexxia: I think the listEeAssets doc is wrong now, sorry I didn't catch it before. The return type is much simpler right?

janakdr: Sorry, what's wrong about it? It's a promise with an array of <id, type>, no?

juliexxia: Ah, I guess we just don't do any actual processing of the type stuff until later so while the result does have those fields, we're not actually using them in listEeAssets so sort of strange to document them in the return type. but maybe not!

janakdr:
Right, we're just passing the EE result straight through. I agree it's a bit strange, but since the EE result isn't documented anywhere, we might as well say it here. Added a line to listEeAssets doc to make this explicit.

PR #376, opened 2019-12-31
Fix bug where we errored if folder already existed

janakdr:
Disasters can share states, and if a disaster folder is created as part of an unsuccessful add (say because Firestore write failed), we want to be able to re-use it.

Also fix bug where we were writing locally before EE and Firestore succeeded, so failures left the page in a bad state. We actually had test coverage for this, but Cypress apparently has a bug.

Main bug here was introduced by #361: we added error handling there, but didn't tolerate "already existing" errors. I had thought in the back of my mind that overwrite=false would prevent errors, but didn't actually test it.

### DIFF
--- a/docs/import/manage_disaster.js
+++ b/docs/import/manage_disaster.js
@@ -362,11 +362,6 @@ async function writeNewDisaster(disasterId, states) {
     return false;
   }
   clearStatus();
-  const currentData = createDisasterData(states);
-  disasterData.set(disasterId, currentData);
-  // We know there are no assets in folder yet.
-  disasterAssets.set(disasterId, Promise.resolve(new Map()));
-
   const eeFolderPromises =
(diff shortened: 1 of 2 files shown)


docs/import/manage_disaster.js
```diff
@@ -421,7 +423,7 @@ function getCreateFolderPromise(dir) {
   return new Promise(
       (resolve, reject) =>
           ee.data.createFolder(dir, false, (result, failure) => {
-            if (failure) {
+            if (failure && !failure.startsWith('Cannot overwrite asset ')) {
```
juliexxia: why only for these failures?

janakdr: The "cannot overwrite asset" failure is benign, because it means the folder is already there, which is what we wanted. Other failures might not be ok (permissions errors, etc.), so we throw. Is that your question?

PR #378, opened 2020-01-02
Add backend for flexible data sources for score asset creation

janakdr:
## Code changes

Add a new function, `create_score_asset#createScoreAssetForFlexibleDisaster`, for a disaster that is not state-based.

The disaster has to have at the very least a poverty asset. If the poverty asset doesn't already have geographies, we need a geography asset as well. If it doesn't already have building counts, we need a building asset as well, or for the damage asset to have all buildings, not just damaged ones. The building asset can either be just damage counts, or have geographies that we compute a histogram over.

Add functionality to filter undamaged buildings out of damage asset. Given that CrowdAI is/was planning to give us damage with all buildings, seems prudent to plan for that.

Refactored the standard score asset creation to share more code with this new codepath. That involved a few more small transformations of the asset, rather than fewer bigger ones, but asset creation time for Harvey didn't change significantly for this PR's `createScoreAssetForStateBasedDisaster` versus master's `createScoreAsset` (actually, it was <30 minutes for both, who knows why so fast).

Also got rid of error checking. It was taking up a lot of screen real estate, and since we have the pre-submission checks, I don't think it's worth it.

## Philosophy

The principle is that we accept the spectrum from an asset that already has everything we need, to the current Census-style data. If the poverty asset already has everything we need, great. It can be missing geographies (as the Census asset is) and buildings. Buildings can come from a Census-like asset. Big violation of this principle is that the poverty asset is expected to have a single "poverty" column that corresponds to percentage. If we want to add "poor households"/"total households" columns to more closely mimic the Census, we can do that in the future. That would also fix the issue in polygon_draw.js, which currently wants to display approximate percentages, for which we need totals in the underlying data to compute weighted averages.

If we ever get actual data to play with, we may find that this setup is unnecessarily flexible or not flexible enough, but it seemed like a reasonable starting point that we can iterate on. My main uncertainty is in allowing the user to specify the columns that correspond to our special columns. We could require the datasets to have the right column names, since I think it's likely that the user will have to do some dataset manipulation even to get it into this "flexible" format, so they might as well rename columns, but maybe not: the Census wouldn't require renaming.

## Future Plans

This function isn't yet connected to anything but unit tests. My plan is, during disaster creation, to allow the user to decide between state-based and "flexible". If they choose flexible, then instead of the current state table, they'll get more of a flow:
1. Specify the poverty asset (maybe just as a string, or from the disaster folder?)
1. Display a field picker for the poverty property
1. If the poverty asset does not have geographies:
    1. Display a new picker for geography asset
    1. Display a poverty asset field picker for the join
    1. Once geography asset is picked, display a geography asset field picker for the join
1. Display three-way choice: building count is in poverty asset, buildings are in separate asset, or use damage for buildings
    - In first case, offer poverty asset field picker for building count field
    - In second case, display buildings asset picker. Once buildings asset is picked, if it doesn't have geographies, display buildings asset field pickers for building count field and district key for join

Damage remains as before, except that we add a damage asset field picker, and a value picker for that field. If "use damage" is specified for buildings, those become mandatory.


janakdr:
About column names, I guess the question is how easy we think it will be for the user to rename columns. If they're comfortable with manipulating feature collections, then great. If not, and they just got a table from somewhere, seems dicier. We could provide an EE script that does the column renames for them, of course, but it could need tweaking depending on the dataset.

docs/import/create_score_asset.js
```diff
@@ -3,67 +3,107 @@ import {getBackupScoreAssetPath, getScoreAssetPath} from '../resources.js';
...
+/**
+ * Given an in-progress score asset, adds in the damage percentage. The damage
+ * asset may include undamaged buildings, in which case we must filter them out
+ * via the `damageLevelsKey` and `noDamageValue` parameters.
+ * @param {ee.FeatureCollection} featureCollection Score asset being created
+ * @param {ee.FeatureCollection} damage Damage collection, which may includes
+ *     undamaged buildings
+ * @param {?string} damageLevelsKey Property in damage collection whose value
+ *     can identify undamaged buildings
+ * @param {?string} noDamageValue Value for `damageLevelsKey` that indicates a
```
juliexxia: Something we definitely need to document - we expect only one damage/undamaged col with all possible states

docs/import/create_score_asset.js
```diff
@@ -95,24 +135,35 @@ function convertToNumber(value) {
...
  */
 function combineWithSnap(feature, snapKey, totalKey) {
   const snapFeature = ee.Feature(feature.get('primary'));
+  const snapPop = convertToNumber(snapFeature.get(snapKey));
+  const totalPop = convertToNumber(ee.Number.parse(snapFeature.get(totalKey)));
+  const badData = null;
+  const snapPercentage = ee.Algorithms.If(
+      totalPop,
+      ee.Algorithms.If(
+          ee.Algorithms.IsEqual(snapPop, null), badData,
```
juliexxia: why using badData in place of null while also using null?

docs/import/create_score_asset.js
```diff
@@ -132,16 +183,43 @@ function combineWithAsset(feature, tag, key) {
...
+    }
+    return result.set(property, null);
+  });
 }
 
+/**
+ * For each {@link ee.Feature}, sets `newProperty` to the value of `property`
+ * for that feature, and sets `property` to null, effectively removing its value
+ * from every feature.
+ *
```
juliexxia: super nit: rm extra line?

docs/import/create_score_asset.js
```diff
@@ -305,26 +320,123 @@ function createScoreAsset(
...
+ *        asset already has a building count. If `useDamageForBuildings` is
+ *        true, see below;
+ * 4. An optional damage asset;
+ *      - If `damageLevelsKey` is present, then undamaged buildings (indicated
+ *        by `noDamageValue`) are filtered out of the damage asset when counting
+ *        damage points in a district;
+ *      - If `useDamageForBuildings` is true, then `damageLevelsKey` must be
+ *        present, and the total number of "damage" points in a district
+ *        (including undamaged ones) is used as the total building count.
+ *
```
juliexxia: super nit: rm extra line?

docs/import/create_score_asset.js
```diff
@@ -305,26 +320,123 @@ function createScoreAsset(
...
     allStatesProcessing = allStatesProcessing.merge(processing);
   }
 
   return backUpAssetAndStartTask(allStatesProcessing);
 }
 
+/**
+ * Performs operation of processing inputs and creating output asset for a
+ * disaster whose data is fairly flexible, with few assumptions. We expect:
+ * 1. A poverty asset;
```
juliexxia: What do you think is the best way to document this flow? I think it's as clear here as text will allow but would be nice to have some kind of flow chart or something (maybe I'm just a visual learner) for this series of if else choices

docs/import/create_score_asset.js
```diff
@@ -467,16 +570,16 @@ function calculateDamage(assetData, setMapBoundsInfo) {
  * This method will go away or be greatly changed if we're using CrowdAI data
  * instead of previously computed building data.
  * @param {string} buildingPath location of buildings asset in EE
- * @param {ee.FeatureCollection} stateGroups Collection with block groups
- * @return {ee.Dictionary} Number of buildings per block group
+ * @param {ee.FeatureCollection} geographies Collection with districts
```
juliexxia: nit: maybe s/districts/geographical areas since districts is an actual census thing

docs/import/create_score_asset.js
janakdr: I think we allow a slightly more permissive structure, although we have to be careful how we document it: if there is a column per damage type, then the user can set the damageLevelsKey to the column for "undamaged", and set noDamageValue to "true" or whatever the sentinel for that column is. So I think this should allow for all our currently known types of damage assets.

docs/import/create_score_asset.js
janakdr: It was my attempt to be a little more document-y: the null in the isEqual is being compared with snapPop. badData is a value that we return. Would a more descriptive name like `nullSnapPercentage` help instead of badData, or is this just a badIdea?

docs/import/create_score_asset.js
janakdr: I think once the front-end is available, that will be the best documentation outside of the doc/code itself. It should walk the user through as they make selections, so the developer can also use it. I don't anticipate, for instance, making this flow explicit to the user. They should just choose the assets they have, and we'll give them the next choice based on that asset, with explanation.

docs/import/create_score_asset.js
janakdr: Can you point me to what "district" means in the census context? I didn't think it was a special word. There used to be an "enumeration district", but I think it's obsolete, and anyway "block group" is the modern version of "enumeration district", so it seems appropriate.

docs/import/create_score_asset.js
juliexxia: Oh great point. Yeah this is going to be a doozy to document. Maybe the best way is to provide a couple different supported forms.

docs/import/create_score_asset.js
juliexxia: Gotcha, no I think this is fine just confirming I wasn't missing anything else going on

docs/import/create_score_asset.js
juliexxia: I think you're right the front end will be the best docs. Might be nice to replicate a flow chart into a doc in case we have front end bugs so it's never the full source of truth on the behavior we expect.

docs/import/create_score_asset.js
juliexxia: Just stuff like this: https://www.census.gov/mycd/

janakdr: Yeah, that seems far enough removed from actual Census definitions that it seems unlikely to confuse. We're going to need a generic word for "block groups" for this work, and "district" seems better than "geographical areas". If you can think of another succinct generic term, let me know!

Issue #172, opened 2019-10-29
Improve error messaging


ruthtalbot:
- Log more precise errors and code paths
- Make errors display as a snackbar on the page

janakdr:
I've done an audit of our codebase, and have PRs out for any remaining catch/throw blocks and createError calls.

PR #380, opened 2020-01-02
Better errors, part 1 of n

janakdr:
* Display an explicit error message if the score asset promise throws.
* Allow `showError` to take an arbitrary object, so that `Error` objects can be logged to the console as objects, allowing inspection.
* Remove now-redundant error handling in draw_table, convert to async for prettiness, and come good on old TODO by adding generic table error handling (although I don't know if we've ever actually encountered it).


docs/run.js
```diff
@@ -100,9 +100,12 @@ function createAndDisplayJoinedData(map, initialTogglesValuesPromise) {
   // clear old listeners
   google.maps.event.removeListener(mapSelectListener);
   google.maps.event.removeListener(featureSelectListener);
+  const dataPromise = resolvedScoreAsset.then(getEePromiseForFeatureCollection);
+  dataPromise.catch(
```
juliexxia: can you attach this to the above line?

janakdr: I don't think so: we want the functions that are called below with dataPromise to see the thrown error, so they don't try to do anything. If we passed the `dataPromise.catch` expression into them, they'd see a successful promise (although one with a value of `undefined`).

If we threw inside the `.catch` block that would prevent them seeing an unsuccessful promise, but I'm not sure it gains anything. I can do that, if you prefer? Will go ahead and submit this after responding to other reviews, but can change in a follow-up.

janakdr:
#172

PR #381, opened 2020-01-02
Clean up some error/code issues around saving polygons

### LINKED ISSUE #172: Improve error messaging

janakdr:
* Show "Saving.../Saved" toast messages, similar to manage_layers/manage_disaster pages
* On error, show reasonable (I hope) snackbar message about what just happened. Not clear if user can actually recover, and how much time we should sink into making this bulletproof, but give it a shot

Closes #172


janakdr:
TODO: if this gets submitted after #382, delete createError method, since this was its last usage.

docs/polygon_draw.js
```diff
@@ -175,22 +206,19 @@ class StoredShapeData {
...
+   *    are done
    */
   finishWriteAndMaybeWriteAgain() {
-    StoredShapeData.pendingWriteCount--;
     const oldState = this.state;
     this.state = StoredShapeData.State.SAVED;
     switch (oldState) {
       case StoredShapeData.State.WRITING:
-        loadingElementFinished(writeWaiterId);
+        this.noteWriteFinished();
```
juliexxia: as discussed offline, maybe do this only when all global writes have been finished

janakdr: Done, and added a lot of complexity for good measure.

docs/polygon_draw.js
```diff
@@ -51,11 +50,31 @@ class StoredShapeData {
...
+        StoredShapeData.errorCount++;
+      }
+    }
+    if (StoredShapeData.errorCount === 0 &&
+        StoredShapeData.pendingWriteCount === 0) {
+      showSavedToast();
+    } else if (StoredShapeData.pendingWriteCount === 0 && success) {
+      showToastMessage(
+          'Latest save succeeded, but there are still ' +
+          StoredShapeData.errorCount + ' feature(s) not saved');
```
juliexxia: At this point, are these every going to be saved? what does recovery look like?

janakdr: The tests give examples. If you encounter a transient Firestore or EE error, modifying the shape and saving could work: if that happens, we drop the error count above, so we'll know that future saves can be displayed as just having succeeded.

juliexxia: Ah helpful - thanks

PR #382, opened 2020-01-03
Get rid of impossible createError call by using existing error code

janakdr:
This should be impossible, but since we have the error-handling code already, just call it.

#172


janakdr:
TODO: if this gets submitted after #381, delete createError method, since this is its last usage.

PR #390, opened 2020-01-04
Migrate Firestore keys to camelCase and start flexible frontend work

janakdr:
I combined these two changes because I was a bit lazy, but hopefully change is not too bad. We add some levels to `assetData`: `stateBasedData` and `flexibleData`. `states` moves inside `assetData.stateBasedData`, since it wasn't used anywhere else. Everything inside `assetData` that wasn't damage-related moves inside `stateBasedData`.

Also add beginning of flexible disaster work: allow user to create it, although nothing can be done with it once it's created. I added the front end, with a warning just in case GD tries to create a disaster while this is in flight.

I didn't move layer-related keys, to avoid conflicting with #365. Will migrate those once that is in.

I processed all our current disaster data in Firestore to create versions in camel case. Should mostly co-exist peacefully together, can delete the old versions when this is submitted. The layers array got moved to layerArray, because `layers` is already in camel case, so it would have overwritten it. Will do a shuffle when we're ready to rename layers fields. If I have time, might push script up so it can be used by others.

Code changes: destructured where possible, used more compact `.assetData` versus `['assetData']` where I saw it. In tests, tried to use expect(foo).to.have.property('bar', val) where possible. Some issues involving deep equality and Chai interfered with that, sadly.

#17


docs/import/create_disaster_lib.js
```diff
@@ -1,44 +1,73 @@
...
-// The default values for column names here are taken from Census American
-// FactFinder and CDC tables.
+const commonAssetDataTemplate = Object.freeze({
+  damageAssetPath: null,
+  damageLevelsKey: null,
+  noDamageValue: null,
+});
+
+const stateBasedDataTemplate = Object.freeze({
+  scoreBoundsCoordinates: null,
```
juliexxia: This PR seems like a good time to do a documentation of the firestore keys in a README somewhere

docs/import/create_disaster_lib.js
```diff
@@ -1,44 +1,73 @@
...
+    paths: {},
+    snapKey,
+    totalKey,
+  },
+  sviAssetPaths: {},
+  sviKey,
+  incomeAssetPaths: {},
+  incomeKey,
+  buildingAssetPaths: {},
+  states: null,
```
juliexxia: what is the difference between
> myKey: null,

and

> myKey,

?

docs/import/create_disaster_lib.js
```diff
@@ -1,44 +1,73 @@
...
-  income_asset_paths: {},
-  income_key: incomeKey,
-  building_asset_paths: {},
-};
-Object.freeze(assetDataTemplate);
+const stateAssetDataTemplate = Object.freeze(
+    {...commonAssetDataTemplate, stateBasedData: stateBasedDataTemplate});
+
+const flexibleAssetData =
+    Object.freeze({...commonAssetDataTemplate, flexibleData: {}});
```
juliexxia: Add TODO for self here? Helpful for me to remember this is code in flux

docs/import/create_disaster_lib.js
juliexxia: We only have scoreBoundCoordinates for state based?

docs/import/create_score_asset.js
```diff
@@ -280,28 +280,29 @@ function setMapBoundsInfo(message) {
...
+    blockGroupAssetPaths,
+    snapData,
+    sviAssetPaths,
+    sviKey,
+    incomeAssetPaths,
+    incomeKey,
+    buildingAssetPaths,
+  } = stateBasedData;
+  const {paths: snapPaths, snapKey, totalKey} = snapData;
+  // If we switch to CrowdAI data, building asset paths will change.
```
juliexxia: This means the structure will change not the value right?

docs/import/manage_disaster.html
```diff
@@ -51,6 +51,15 @@
     <form action="" id="new-disaster-form">
       <label for="name">Disaster Name: </label><input id="name"><label for="name"><i> ex: harvey</i></label><br>
       <label for="year">Disaster Year: </label><input id="year"><label for="year"><i> ex: 2017</i></label><br>
+      <div>
+      Will you be using U.S. Census data for this disaster? If not, you will have to provide
```
juliexxia: Is the real distinction here is the disaster US state-based or not? Like there is no situation where a disaster hits one of the 50 US states and the user shouldn't take the traditional approach right? It might be confusing for people who aren't familiar with the available data that 50 states disaster === US Census Data.

docs/import/manage_disaster.html
```diff
@@ -51,6 +51,15 @@
...
+      poverty/geography/buildings datasets in appropriate formats.<br/>
+      <input type="radio" name="disaster-type" id="disaster-type-census" value="census" checked onclick="document.getElementById('states-select-div').hidden = false">
```
juliexxia: These radios might not need values

docs/import/manage_disaster.html
```diff
@@ -51,6 +51,15 @@
...
+      <label for="disaster-type-census">U.S. Census data available</label>
+      <input type="radio" name="disaster-type" id="disaster-type-flexible" value="flexible" onclick="document.getElementById('states-select-div').hidden = true">
+      <label for="disaster-type-flexible"><b>NOT YET FUNCTIONAL! DO NOT USE</b> No U.S. Census data available: I will provide appropriate datasets. </label>
```
juliexxia: nit: maybe semicolon instead of colon? and maaaybbeeee "I (the user) will..." or something to make it clear that we (the developers) are not the subject here

docs/import/manage_disaster.js
```diff
@@ -331,10 +335,16 @@ function deleteDisaster() {
...
-  const states = $('#states').val();
+  let states = $('#states').val();
 
-  if (!year || !name || !states) {
-    setStatus('Error: Disaster name, year, and states are required.');
+  if (!year || !name) {
+    setStatus('Error: Disaster name and year are required.');
+    return Promise.resolve(false);
+  }
+  if ($('#disaster-type-flexible').val()) {
```
juliexxia: does is(':checked') work here?

docs/import/create_disaster_lib.js
janakdr: As you know, I'm a fan of keeping docs and code together when possible. I can add inline comments to the fields in this file, would you like that? It would help to understand who the intended audience is. My impression is that someone who needs to know what a Firestore field means is the same as someone who needs to know what some global variable in our code means, and they should be documented the same way. Do you envision a different audience?

janakdr: mistake. Fixed in Firestore, but not code. Thanks for catching!

docs/import/create_disaster_lib.js
janakdr: JavaScript destructuring: the second one means `'myKey': myKey,`

docs/import/create_disaster_lib.js
janakdr: There's no TODO here. I'm planning to add the fields as they get selected by the user, since they don't all need to be set. That's in contrast to above, where by the end we expect all the state-based fields to be set.

docs/import/create_score_asset.js
janakdr: Removed. I think this is taken care of by the TODO below about filtering buildings out from damage, and Hans seems happy with never using CrowdAI buildings data.

docs/import/manage_disaster.html
janakdr: Yes, but they might also use Census data for a U.S. territory, or they might not. Census data seemed like the actual criterion. Happy to rephrase. Do you prefer "Are you focused on this disaster's impact on U.S. states and/or territories with Census data available?" Seems unwieldy to me. Other suggestions?

docs/import/manage_disaster.html
janakdr: Done for semicolon. I don't think anyone will think "I" doesn't mean them except for us :)

docs/import/manage_disaster.js
janakdr: Done, thanks.

docs/import/manage_disaster.html
janakdr: Done, thanks.

docs/import/manage_disaster.html
juliexxia: Yeah, honestly why i didn't leave a suggestion was because I couldn't think of something quippy. Maybe the answer is to link to a help that leads to a long section @ruthtalbot can write about what data we know exists at this point?

docs/import/create_disaster_lib.js
juliexxia: I'm good with that! All the keys are in this file right?

docs/import/create_disaster_lib.js
juliexxia: In that case - can we document all possible fields here in comments the way we're discussing documenting all the recorded fields for state based

docs/import/manage_disaster.html
ruthtalbot: I believe we have working census data for PR but not the other territories. Can add documentation around this in the Data Source guide. Is it specifically US Census SNAP data we need here? 

Added in this heading here which I will fill in with background if you want to link to it: https://docs.google.com/document/d/1AyjEM5-z-JRII3yJ2vx1awmFLBzWilXw9gIuhMWItYI/edit?ts=5da9e18d#heading=h.e45lmrvbllcp

janakdr: SNAP data and TIGER shapefiles, I think.

docs/import/create_disaster_lib.js
janakdr: Done.

docs/import/create_disaster_lib.js
janakdr: Done. The `layers` keys aren't in here, neither are mapBounds, etc., but that can be for another day.

docs/import/manage_disaster.html
janakdr: I added a link to the general help doc here. @ruthtalbot I think we need a higher-level discussion of when you would want a Census-based vs flexible disaster. Left a TODO for you here to add that.

PR #391, opened 2020-01-04
Add script used to migrate keys to camel case

janakdr:
Ran this in order to enable #390. Seems worth having in the repo for future needs.

### DIFF
--- a/admin_utils/process_firestore.js
+++ b/admin_utils/process_firestore.js
@@ -65,7 +65,7 @@ async function main() {
 }
 
 /**
- * Recursively process `data` and return a new object with all keys in camel
+ * Recursively processes `data` and returns a new object with all keys in camel
  * case, with a few exceptions. We transform `layers` to `layerArray` because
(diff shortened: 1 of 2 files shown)


admin_utils/process_firestore.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {Object} data Data coming from Firestore
+ * @param {Set<string>} mangledKeys Keys whose values were changed
+ * @return {Object} Processed data
+ */
+function camelCase(data, mangledKeys) {
+  // We don't want to modify any primitives.
+  if (!data) {
+    return data;
+  }
+  if (typeof (data) !== 'object') {
```
juliexxia: Will this clause not take care of the previous one?

janakdr: Very sadly, no. typeof(null) === 'object'.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null

PR #392, opened 2020-01-04
Try to deflake new test

janakdr:
Failed two times in a row for me, but pretty hard to reproduce. Think this fixed it, we'll see.


janakdr:
Going to wait to submit until I see it flake one more time...

janakdr:
Ok, saw a flake, and not with my pending PR that I was afraid was somehow the cause. Submitting.

PR #393, opened 2020-01-04
Revive image upload page

janakdr:
Verified that it's working for basic cases: successfully uploads an image file to GCS then imports to EE, avoids re-uploading to GCS if already present, deletes from GCS if already in EE, tells you to delete if already in EE.

Main change besides those necessary to resurrect is that we now create the image collection in the current disaster folder, and similarly create the GCS files under a path that includes the current disaster.

EE seems to now complain if the GCS URI doesn't include an extension, so we now only transform illegal characters for the EE asset, not the GCS file.

Made a minimal effort to show errors better and clean up a bit.

Still needs tests, but I think worth bringing in as is. Will think about how to test: probably have to mock everything but ee listing out, so will just get very basic logic testing.

#173


docs/import/upload_image_collection.html
```diff
@@ -0,0 +1,44 @@
...
+  <script src="../external_libs/firebase-auth-7.6.1.js"></script>
+
+  <script type="module" src="./upload_image_collection_startup.js"></script>
+
+  <!-- Load our styles. -->
+  <link rel="stylesheet" href="../style.css">
+</head>
+<body>
+
+<div id="navbar"></div>
```
juliexxia: I don't see a navbar showing up on the page - not supposed to be hooked up yet?

Also don't see a header though there appears to be space for one.

docs/import/upload_image_collection.html
```diff
@@ -0,0 +1,44 @@
...
+  <!-- TODO(janakr): This is pinned to a particular version, but we should be targeting the live
+       version, whatever it is. -->
+  <script src="https://rawcdn.githack.com/google/earthengine-api/3bb86bfc4f3d9eed98220f3d225b414982915b86/javascript/build/ee_api_js_debug.js"></script>
+  <script src="https://apis.google.com/js/api.js"></script>
+
+  <script src="../external_libs/firebase-app-7.6.1.js"></script>
+  <script src="../external_libs/firebase-firestore-7.6.1.js"></script>
+  <script src="../external_libs/firebase-auth-7.6.1.js"></script>
+
+  <script type="module" src="./upload_image_collection_startup.js"></script>
```
juliexxia: Get this error when I visit the page: 

http://localhost:8080/import/upload_image_collection_startup.js net::ERR_ABORTED 404 (Not Found)

supposed to be upload_image_collection.js?

docs/import/upload_image_collection.js
```diff
@@ -1,18 +1,18 @@
...
 
-export {taskAccumulator};
+import {listEeAssets} from './ee_utils.js';
 
-const earthEngineAssetBase = gdEePathPrefix + getDisaster();
-const earthEnginePrefix = eeLegacyPathPrefix + earthEngineAssetBase;
+export {setUpAllHeaders};
 
-const BUCKET = 'givedirectly.appspot.com';
+const EARTH_ENGINE_PREFIX = eeLegacyPathPrefix;
```
juliexxia: why reinitialize?

docs/import/upload_image_collection.js
```diff
@@ -178,35 +164,41 @@ function uploadFileToGCS(name, contents, collectionName, callback) {
...
-        resultDiv.innerHTML += '<br>Error processing ' + name + ': ' + err;
-      });
+  let bucket;
+  try {
+    const result = await fetch(URL, request);
+    const json = await result.json();
+    if (result.status !== 200) {
+      throw new Error('Unable to fetch from GCS: ' + json.error.message);
+    }
+    ({bucket} = json);
```
juliexxia: what are these outer parens for ?

docs/import/upload_image_collection.js
```diff
@@ -178,35 +164,41 @@ function uploadFileToGCS(name, contents, collectionName, callback) {
...
- * Checks to see if asset given by collectionName is present. If not, creates it
- * and makes it world-readable.
+ * Checks to see if asset given by `qualifiedName` is present. If not, creates
+ * it and makes it world-readable.
  *
- * @param {string} collectionName Must contain only legal EE characters
- * @return {Promise<undefined>} Promise to wait for operation completion on
+ * @param {string} qualifiedName Must contain only legal EE characters. Like
+ *     '2017-harvey/noaa-images'
+ * @return {Promise<void>} Promise to wait for operation completion on
```
juliexxia: I know this isn't new, but having a difficult time parsing this return description

docs/import/upload_image_collection.html
janakdr: Sorry, forgot to push file. Should be there now.

docs/import/upload_image_collection.html
janakdr: Because of missing startup file. Should be fixed.

docs/import/upload_image_collection.js
janakdr: removed.

docs/import/upload_image_collection.js
janakdr: Necessary for destructuring. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment "Assignment without Declaration" section.

docs/import/upload_image_collection.js
janakdr: Rewritten. Better?

docs/import/upload_image_collection_startup.js
```diff
@@ -0,0 +1,19 @@
+import {trackEeAndFirebase} from '../authenticate.js';
+import {loadNavbarWithPicker} from '../navbar.js';
+import {TaskAccumulator} from '../task_accumulator.js';
+
```
juliexxia: nit: extra \n

janakdr:
Sure, done.

PR #400, opened 2020-01-06
Document that our tests use live Harvey data

janakdr:


INITIAL_SETUP.md
```diff
@@ -125,6 +125,27 @@ To set up a brand-new system with new Google accounts:
...
 
+#### Set up test data
+ 
+Data from Hurricane Harvey (2017) must be present, since our integration tests
+assert against it.
+
+-  As the Admin account (currently `gd-earthengine-user@givedirectly.org`), go
+   to the
+   [manage disaster](https://givedirectly.github.io/Google-Partnership/import/manage_disaster.html)
+   page and add a disaster named `harvey` in `2017`, with Texas the affected
```
ruthtalbot: nit: "Texas as the affected.." ?

INITIAL_SETUP.md
```diff
@@ -125,6 +125,27 @@ To set up a brand-new system with new Google accounts:
...
+   state, if it does not already exist.
+
+-  Follow [instructions](https://docs.google.com/document/d/1WnjHIexMnlVkn5lqnzYlBVg4aWxh2wVDZI2DBAu-TGI/edit?disco=AAAAEE2MdUU&ts=5e1368bd&usp_dm=true)
```
ruthtalbot: Feels a little strange to be telling them to create harvey but also assuming we have a folder that already exists for harvey that stores this damage data -- is the create step just there in case firebase gets wiped?

INITIAL_SETUP.md
janakdr: I think this is grammatical too :)

INITIAL_SETUP.md
janakdr: Or if a new user is setting this up, but the GD setup continues to exist.

Issue #333, opened 2019-12-18
Standardize manage_* status divs


janakdr:

janakdr:
And/or improve messaging during page load, etc.

janakdr:
@RamV13 FYI.

We could also do a better job of not jumping around during page load. Currently, I think we only create the state pickers in manage_disaster.html once all Firestore *and* EE calls have completed. That's not necessary, we could render the table with disabled selections before the EE calls are done, and swap out the placeholders. Or we could just have a prominent progress indicator over the place the future table is going to be.

PR #403, opened 2020-01-07
Unifying styles for status divs on manage pages

RamV13:
Starts off the work for #333


RamV13:
Ah I see yup might make sense to just delete this altogether later on then.

PR #407, opened 2020-01-09
Adding icons to popup actions

RamV13:


docs/popup.js
```diff
@@ -168,6 +174,9 @@ function setUpPopup() {
       const closeButton = document.createElement('button');
       closeButton.className = 'popup-button';
       closeButton.innerHTML = 'close';
+      const closeIcon = document.createElement('i');
```
janakdr: Would you mind adding an icon for the save button, above?

RamV13: Ah missed that thanks

PR #414, opened 2020-01-10
Allow for flexible columns when computing score

### LINKED ISSUE #322: Make damage percentage undefined ('-'?) instead of 0 for no-damage score asset

janakdr:
* Write important parameters used to create score asset to Firestore. Move the old parameters if we're moving the asset, so it all stays in sync.
* Make score computation use parameters set when computing score asset, instead of hard-coded ones.
* We now know whether damage was actually used when creating score asset, versus just whether it exists now.
* Order columns according to some criteria (score -> poverty -> damage), and display in pop-up in that order as well. Also show "(data unavailable)" if a feature is missing data that other features have.
* Add a bunch of parameters to `assetData.flexibleData`, in preparation for follow-up PR. This allows user to customize the columns that are used for various things, and doesn't rename the columns, so that the user can see the original column names on the map, rather than something generic. `GEOID` is the only one renamed, since it's never shown to the user.
* Support filtering out no-damage buildings from damage asset even in state-based score asset creation. Currently there's no way for user to specify no-damage buildings, so doesn't matter.
* Change `firestore_document#readDisasterDocument` to return the actual data, since it's supposed to be read-only anyway.
* async/await a few functions I came across.
* Add test coverage for failure to start an EE task, which I ran across in real life.

#305 #17 Closes #322

### DIFF
--- a/docs/click_feature.js
+++ b/docs/click_feature.js
@@ -91,8 +91,10 @@ function createHtmlForPopup(feature, rowData, scoreParameters, columns) {
     if (value && column.endsWith(' PERCENTAGE')) {
       value = parseFloat(value).toFixed(3);
     }
-    property.innerText = column + ': ' +
-        (value !== null && value !== undefined ? value : '(data unavailable)');
+    property.innerHTML = column + ': ' +
+        (value !== null && value !== undefined ?
+             value :
+             '<span class="data-unavailable-span">(data unavailable)</span>');
     properties.appendChild(property);
   }
   div.appendChild(heading);
--- a/docs/click_feature.js
+++ b/docs/click_feature.js
@@ -1,6 +1,6 @@
 import {showError} from './error.js';
 import {currentFeatures, highlightFeatures} from './highlight_features.js';
-import {blockGroupTag, geoidTag, scoreTag} from './property_names.js';
+import {geoidTag, scoreTag} from './property_names.js';
 
 export {clickFeature, selectHighlightedFeatures};
 
@@ -14,8 +14,11 @@ export {clickFeature, selectHighlightedFeatures};
  * @param {string|ee.FeatureCollection} featuresAsset asset (path) of features
  *     which could be clicked
  * @param {Function} tableSelector See drawTable
+ * @param {ScoreParameters} scoreParameters Needed for `districtDescriptionKey`
+ * @param {Array<EeColumn>} columns Properties to show on click
  */
-function clickFeature(lng, lat, map, featuresAsset, tableSelector, scoreParameters, columns) {
+function clickFeature(
+    lng, lat, map, featuresAsset, tableSelector, scoreParameters, columns) {
   const point = ee.Geometry.Point(lng, lat);
   const blockGroups = ee.FeatureCollection(featuresAsset).filterBounds(point);
   const selected = blockGroups.first();
@@ -37,7 +40,8 @@ function clickFeature(lng, lat, map, featuresAsset, tableSelector, scoreParamete
       highlightFeatures([feature], map);
       const rowData = tableSelector([geoid]);
       const infoWindow = new google.maps.InfoWindow();
-      infoWindow.setContent(createHtmlForPopup(feature, rowData, scoreParameters, columns));
+      infoWindow.setContent(
+          createHtmlForPopup(feature, rowData, scoreParameters, columns));
       const borderPoint = feature.geometry.coordinates[0][0];
       infoWindow.setPosition(
           new google.maps.LatLng({lat: borderPoint[1], lng: borderPoint[0]}));
@@ -57,6 +61,8 @@ const HIDDEN_PROPERTIES = Object.freeze(new Set([
  * @param {Feature} feature post-evaluate JSON feature
  * @param {?Array<string>} rowData Data from selected row in table for this
  *     feature, if found
+ * @param {ScoreParameters} scoreParameters
+ * @param {Array<EeColumn>} columns
  * @return {HTMLDivElement} Div with information
  */
 function createHtmlForPopup(feature, rowData, scoreParameters, columns) {
@@ -85,7 +91,8 @@ function createHtmlForPopup(feature, rowData, scoreParameters, columns) {
     if (value && column.endsWith(' PERCENTAGE')) {
       value = parseFloat(value).toFixed(3);
     }
-    property.innerText = column + ': ' + (value !== null && value !== undefined ? value : '(data unavailable)');
+    property.innerText = column + ': ' +
+        (value !== null && value !== undefined ? value : '(data unavailable)');
     properties.appendChild(property);
   }
   div.appendChild(heading);
--- a/docs/click_feature.js
+++ b/docs/click_feature.js
@@ -15,7 +15,7 @@ export {clickFeature, selectHighlightedFeatures};
  *     which could be clicked
  * @param {Function} tableSelector See drawTable
  */
-function clickFeature(lng, lat, map, featuresAsset, tableSelector) {
+function clickFeature(lng, lat, map, featuresAsset, tableSelector, scoreParameters, columns) {
   const point = ee.Geometry.Point(lng, lat);
   const blockGroups = ee.FeatureCollection(featuresAsset).filterBounds(point);
   const selected = blockGroups.first();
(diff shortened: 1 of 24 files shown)


docs/import/create_disaster_lib.js
```diff
@@ -114,34 +116,72 @@ const stateBasedDataTemplate = Object.freeze({
...
+ *     Asset can contain geometries (in which case each geometry corresponds to
+ *     a building), or just have rows, in which case each row has a building
+ *     count that will be joined to the poverty asset. `buildingGeoid` will
+ *     be the joining column and `buildingKey` the column with the count.
+ * @param {number} POVERTY Indicates that poverty asset already has a building
+ *     count per district. `buildingKey` will identify the column.
+ * @param {number} DAMAGE Indicates that damage asset contains geometries for
+ *     all buildings, not just damaged ones, so the building count can be taken
+ *     from it. If specified, `flexibleData` fields `noDamageKey` and
+ *     `noDamageValue` must be set if `assetData.damageAssetPath` is. (If this
```
juliexxia: Maybe I'll see this in code below, so DAMAGE is the default setting for if there's no damage asset? Maybe just create a 4th setting for clarity?

docs/import/create_disaster_lib.js
```diff
@@ -114,34 +116,72 @@ const stateBasedDataTemplate = Object.freeze({
...
+
 /**
  * Has data for a "flexible" (non-Census-based) disaster.
  * @type {Readonly<Object>}
  * @property {EeFC} povertyPath Contains poverty data. May have geometries, in
  *     which case `geographyPath` is not used. May have building counts, in
  *     which case `buildingPath` is not used. All columns from this asset end up
- *     in final score asset. Must have {@link povertyPercentageTag} column.
+ *     in final score asset.
+ * @property {EeColumn} povertyGeoid Column of `povertyPath` that contains
```
juliexxia: I wonder if it'd be helpful to append these with '(e.g. 'Geoid2')' or something like that so that if there's confusing people can draw comparisons to the census data. Idk.

docs/import/create_score_asset.js
```diff
@@ -1,21 +1,30 @@
...
   createScoreAssetForFlexibleDisaster,
   createScoreAssetForStateBasedDisaster,
   setStatus,
 };
+
 // For testing.
 export {backUpAssetAndStartTask};
 
+// State-based tags.
+
```
juliexxia: nit: extra \n

docs/import/create_score_asset.js
```diff
@@ -385,14 +404,21 @@ function createScoreAssetForFlexibleDisaster(
       calculateDamage(assetData, setMapBoundsInfoFunction);
   const {flexibleData} = assetData;
   let processing = ee.FeatureCollection(flexibleData.povertyPath);
-  const {povertyGeoid, geographyPath, buildingPath, buildingKey} = flexibleData;
-  const {useDamageForBuildings} = assetData;
+  const {povertyGeoid, povertyHasGeometry, buildingSource} = flexibleData;
+  let {buildingKey} = flexibleData;
+  if (!buildingKey) {
```
juliexxia: when does this clause trigger?

docs/import/create_score_asset.js
```diff
@@ -401,99 +427,132 @@ function createScoreAssetForFlexibleDisaster(
...
-      });
+  let renamed = false;
+  try {
+    await renameAssetAsPromise(scoreAssetPath, oldScoreAssetPath);
+    renamed = true;
+  } catch (renameErr) {
+    // These checks aren't perfect detection, but best we can do.
+    if (renameErr.includes('does not exist')) {
+    } else if (renameErr.includes('Cannot overwrite asset')) {
+      // Delete and try again.
```
juliexxia: nit: mind changing this to Delete existing backup asset and try again

docs/import/create_score_asset.js
```diff
@@ -616,3 +675,22 @@ function displayGeoNumbers(latLngs) {
...
+ * @return {string}
+ */
+function getDamageAssetPath(assetData) {
+  const {damageAssetPath} = assetData;
+  if (!damageAssetPath || typeof (damageAssetPath) === 'string') {
+    return damageAssetPath;
+  }
+  if (inProduction()) {
+    // Should never happen.
+    setStatus('Error specifying damage asset path ' + damageAssetPath);
```
juliexxia: Does this happen in testing or something? Wondering how this came about

docs/process_joined_data.js
```diff
@@ -55,42 +58,68 @@ const ALWAYS_PRESENT_KEYS = Object.freeze([geoidTag, blockGroupTag, scoreTag]);
...
+ *     {@link damageTag} is next, followed by `buildingKey` from
+ *     {@link ScoreParameters}. If damage is absent, `buildingKey` is last.
+ *     Remaining columns are sorted by order encountered in the feature, which
+ *     generally means alphabetically, since EarthEngine always sorts them.
  */
 function processJoinedData(
-    dataPromise, scalingFactor, initialTogglesValuesPromise) {
-  return Promise.all([dataPromise, initialTogglesValuesPromise])
+    dataPromise, scalingFactor, scoreComputationParameters) {
+  return Promise.all([dataPromise, scoreComputationParameters])
```
juliexxia: super nit: I think we've been sticking to appending these vars with Promise when they're promises

docs/run.js
```diff
@@ -77,11 +77,32 @@ function resolveScoreAsset() {
 function run(map, firebaseAuthPromise, disasterMetadataPromise) {
   setMapToDrawLayersOn(map);
   resolveScoreAsset();
+  disasterMetadataPromise = massageDisasterMetadataPromiseWhenUsingBackupAsset(
+      disasterMetadataPromise);
   const initialTogglesValuesPromise =
-      setUpToggles(disasterMetadataPromise, map);
+      setUpScoreComputationParameters(disasterMetadataPromise, map);
```
juliexxia: do you still want to pass around as initialTogglesValuesPromise?

docs/process_joined_data.js
```diff
@@ -55,42 +58,68 @@ const ALWAYS_PRESENT_KEYS = Object.freeze([geoidTag, blockGroupTag, scoreTag]);
...
+                  povertyRateKey,
+                  districtDescriptionKey,
+                  damageAssetPath,
+                },
               },
             ]) => {
-        const columnsFound = new Set(ALWAYS_PRESENT_KEYS);
+        const hasDamage = !!damageAssetPath;
+        const columnsFound = new Set([geoidTag]);
+        columnsFound.add(districtDescriptionKey);
```
juliexxia: why not add all these at once?

docs/run.js
```diff
@@ -91,11 +112,11 @@ let featureSelectListener = null;
  * Creates the score overlay and draws the table
  *
  * @param {google.maps.Map} map main map
- * @param {Promise<Array<number>>} initialTogglesValuesPromise promise that
- *     returns the poverty and damage thresholds and the poverty weight (from
- *     which the damage weight is derived).
+ * @param {Promise<ScoreComputationParameters>} scoreParametersPromise Will
+ *     resolve once score asset name is known and Firestore fetch done
+ * @return {Promise<void>} Promise that is done when score layer all done
```
juliexxia: can you specify all done with what? loading?

docs/update.js
```diff
@@ -63,39 +65,49 @@ function getUpdatedValue(toggle) {
...
-      toggles.set(damageThresholdKey, 0.5);
-      toggles.set(povertyWeightKey, 0.5);
-    }
-    createToggles(map);
-    return getToggleValues();
-  });
+async function setUpScoreComputationParameters(disasterMetadataPromise, map) {
+  ({scoreAssetCreationParameters} = await disasterMetadataPromise);
+  if (!!scoreAssetCreationParameters.damageAssetPath) {
+    toggles.set(damageThresholdKey, 0.5);
```
juliexxia: @ruthtalbot  was wondering if we should set these lower by default, for the case where a journalist looks at it. 0.5 is pretty high threshold and it might make nothing show up at first glance.

ruthtalbot: I don't think it's a huge deal, but I can see a case where someone lands on the map and doesn't think there's any Scored areas because there are so few (I thought that for a minute with Matthew).

docs/process_joined_data.js
janakdr: Thanks, done

docs/run.js
janakdr: Expanded.

docs/update.js
janakdr: Will leave that for a separate PR

docs/click_feature.js
```diff
@@ -91,8 +91,10 @@ function createHtmlForPopup(feature, rowData, scoreParameters, columns) {
     if (value && column.endsWith(' PERCENTAGE')) {
       value = parseFloat(value).toFixed(3);
     }
-    property.innerText = column + ': ' +
-        (value !== null && value !== undefined ? value : '(data unavailable)');
+    property.innerHTML = column + ': ' +
+        (value !== null && value !== undefined ?
+             value :
+             '<span class="data-unavailable-span">(data unavailable)</span>');
```
juliexxia: could just do the styling in the html since we're already having some html in here. Up to you!

janakdr: Prefer to put it in separate file in case we switch here to creating the span via `document.createElement`, or we want to re-style the span.

PR #415, opened 2020-01-11
Adding loading overlays

### LINKED ISSUE #333: Standardize manage_* status divs

RamV13:
Resolves #333

### DIFF
--- a/docs/import/manage_disaster_flexible.js
+++ b/docs/import/manage_disaster_flexible.js
@@ -1,5 +1,7 @@
 import {showError} from '../error.js';
+import {loadingElementFinished} from '../loading.js';
 import {getDisaster} from '../resources.js';
+
 import {BuildingSource} from './create_disaster_lib.js';
 import {getDisasterAssetsFromEe} from './list_ee_assets.js';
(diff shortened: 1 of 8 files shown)


RamV13:
Ah thanks I've extended the loading overlay to cover the flexible data.

I'm not sure how worthwhile it would be to add the loading overlays on the "damage" asset and the layer asset as well given that these aren't tables but @ruthtalbot can probably comment on this.

janakdr:
Hmm, can you say more about why the fact that it's a table is relevant? The
table is just to present the select elements. The damage selector is
similar to the selects that are in the table.

The table used to be very slow to even take shape, so it would make more
sense to show that it was loading, but now it's initialized quickly, and
the main work is waiting for the select values to populate.

RamV13:
I was thinking the loading overlay is more important for the table because of the time it takes to take shape, although like you mentioned it's not that slow anymore.

It seems like the "pending" we have in the selectors is pretty useful already. I'm not sure how useful an additional indicator like the loading overlay would be on these selectors.

docs/import/manage_disaster_flexible.js
```diff
@@ -1,5 +1,7 @@
 import {showError} from '../error.js';
+import {loadingElementFinished} from '../loading.js';
 import {getDisaster} from '../resources.js';
+
```
janakdr: nit: remove blank line

PR #416, opened 2020-01-13
Add front end for flexible disasters

janakdr:
Also add inputs to state-based damage asset: can now filter out no-damage buildings, for CrowdAI use.

Main "flexible" additions are to manage_disaster_flexible.js. See file-level comment there for complexities: in short, since value of one select can influence other values and requirements, change handlers are more complex and validation has to become asynchronous.

Because manage_disaster.js was getting unwieldy, factored it into four (now five) files: manage_disaster.js, which handles the overall setup, manage_disaster_add_delete.js, manage_disaster_base.js, which has common functionality, and manage_disaster_state_based.js, which has the state-based-specific functionality that was all that manage_disaster.js used to support.

Add "pending" selects that show up as soon as we have enough information to display page structure (when Firestore data is retrieved on page load, and instantly on disaster change). This should help inform users what's going on much better. We use a tactic juliexxia@ originally suggested and I was against, but I think now it's the best way: having a single select and replacing its contents with 'pending...' when data is not ready. Having a separate select causes too many issues around tracking it and showing/hiding it.

Harmonize input ids: they are now deterministically derived from property paths. This means that we can stop passing selects and ids around, and just pass property paths, getting selects if necessary.

Change `assetData.flexibleData` structure a bit: in order to keep select ids unique, stop overloading `buildingKey` to carry the value if it's coming from the poverty asset or the buildings asset. Instead, have a `buildingCountKey` and a `povertyBuildingKey`.

manage_disaster_add_delete.js should be a pretty straight refactoring: comparing side by side with manage_disaster.js@master is probably useful. manage_disaster_state_based.js has more changes (mainly to make the "pending" selects show up right away and to deal with the harmonized ids), but should be largely the same. manage_disaster_base.js is changed extensively enough that while manage_disaster.js@master is the basis for a lot, it might not help to compare except if something is confusing. I tried to make it as self-contained as possible, but it does need to use some functions from manage_disaster_flexible.js, unfortunately.


docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+  setOptionsForSelect(columns, POVERTY_BUILDINGS_PATH);
+}
+
+/**
+ * Initializes `assetKey` div: create pending asset select and column selects.
+ * @param {ScoreInputType} assetKey
+ * @return {JQuery<HTMLSelectElement>} Asset select
+ */
+function createSelectAndColumns(assetKey) {
+  const scoreInputData = componentsData[assetKey];
```
juliexxia: could destructure here?

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+    return setInitialColumnValues('geography');
+  }
+}
+
+/**
+ * Initializes poverty-related display. In addition to doing the analog of what
+ * {@link initializeGeography} does, there are two complications: geography div
+ * visibility is based on poverty asset having geometries, and we must
+ * initialize a column select (for the building-count column) in a different
+ * div from the standard poverty div.
```
juliexxia: why do you need a different div for the building-count column select

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+}
+
+/**
+ * Handles basic operation of an asset selection change: sets relevant columns
+ * to 'pending', writes to Firestore, and returns columns of newly chosen asset.
+ * Calls {@link startPending} via
+ * {@link startPendingWriteSelectAndGetPropertyNames}, so callers must call
+ * {@link finishPending}!
+ * @param {ScoreInputType} key
+ * @return {Promise<?Array<EeColumn>>} Null if selection no longer current,
```
juliexxia: Can you add description for when it's not null (or maybe just link to start startPendingWriteSelectAndGetPropertyNames is enough). Though I had to trace it down quite a few methods so might be worth it to just describe here

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+}
+
+/**
+ * @param {EeFC} assetName
+ * @return {boolean} If buildings asset has geometry or unset. Done instantly
+ */
+async function buildingsHasGeometry(assetName) {
+  if (!assetName) {
+    // Assume geometry when they finally choose a buildings asset.
+    return true;
```
juliexxia: slightly confused by this early return

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+ *     contain columns of damage asset. Created via {@link verifyAsset} or
+ *     {@link startPendingWriteSelectAndGetPropertyNames}
+ * @param {?EeFC} damageAsset Value of damage asset, from page
+ * @return {Promise<void>}
+ */
+async function displayDamageRelatedElements(propertyNamesPromise, damageAsset) {
+  setNoDamageColumnAndValue(null);
+  showHideDamageAndMapDivs(!!damageAsset);
+  try {
+    const propertyNames = await propertyNamesPromise;
```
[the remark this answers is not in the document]
janakdr: No, but hard to guarantee, since we're calling out to both Firestore and EE.

If you're asking because of the `try {}` block, it's helpful to also execute the `finally` clause on early returns.

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+ * Gets assets for the current disaster from EarthEngine, then sets those
+ * disasters as options for the select element given by `propertyPath`.
+ * @param {PropertyPath} propertyPath
+ * @param {boolean} enableAllFeatureCollections True to enable all Feature
+ *     Collections, false to leave the defaults (geometry required).
+ * @return {Promise<boolean>} True if successful, false if disaster changed
+ *     while waiting for asset listing, in which case caller should abort
+ */
+async function getAssetsAndSetOptionsForSelect(
+    propertyPath, enableAllFeatureCollections = true) {
```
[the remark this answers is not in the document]
janakdr: non-feature collections were already disabled. Expanded comment in getDisasterAssetsFromEe to say that explicitly. This re-enables ones without geometries.

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+  // Same promise as waited on for many assets. Since getDisasterAssetsFromEe
```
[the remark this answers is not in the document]
janakdr: Tried to rephrase.

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+          .on('change',
+              () => handleAssetDataChange(select.val(), propertyPath));
+  return select;
+}
+
+//                          Save-related functions.
+
+/**
+ * Called on changes to select-type inputs with no required columns, but
+ * "cascading" effects: other inputs whose values depend on this asset's
```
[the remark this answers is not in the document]
janakdr: Tried to rephrase. Any better?

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+function handleAssetDataChange(val, propertyPath) {
+  // We want to change the value, which means we have to write an expression
+  // like "parent[prop] = val". To obtain the parent object, we just follow the
+  // same path as the child's, but stop one property short. That last property
+  // is then the "prop" in the expression above.
+  const parentProperty = getStoredValueFromPath(propertyPath.slice(0, -1));
+  parentProperty[propertyPath[propertyPath.length - 1]] =
+      val !== '' ? val : null;
+  if (isFlexible()) {
+    // This will short-circuit if there are any pending checks, which will be
```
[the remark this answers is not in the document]
janakdr: Expanded. Better?

docs/import/manage_disaster_base.js
```diff
@@ -0,0 +1,805 @@
...
+ * @return {JQuery<HTMLLIElement>} List item will have label and select
+ */
+function createSelectListItemFromColumnInfo(columnInfo) {
+  return createListItem(columnInfo)
+      .append(createSelectWithSimpleWriteOnChange(columnInfo.path));
+}
+
+/**
+ * Creates an `li` for `columnInfo`: Span with Label followed by optional
+ *     (explanation). Explanation has span with id for later modification.
```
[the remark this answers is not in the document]
janakdr: trying to be cute: that's how it's displayed. Also why "Label" is capitalized. This look any better?

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+  return addColumnArrayErrorsToMessage(
+      message, componentsData[key].columns, key);
+}
+
+/**
+ * Checks columns in `columnInfos` to make sure they all have values. If not,
+ * appends errors to `message`.
+ * @param {string} message Current error message
+ * @param {Array<ColumnInfo>} columnInfos
+ * @param {string|ScoreInputType} scoreInputType String to display in error
```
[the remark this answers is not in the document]
janakdr: Said when

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+}
+
+/**
+ * Checks all `columnInfos` for values, concatenates missing ones' labels.
+ * @param {Array<ColumnInfo>} columnInfos
+ * @return {string}
+ */
+function validateColumnArray(columnInfos) {
+  return columnInfos.map(validateColumnPathHasValue)
+      .filter((c) => c)
```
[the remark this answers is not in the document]
janakdr: Added comment.

docs/import/manage_disaster_flexible.js
```diff
@@ -0,0 +1,740 @@
...
+      .join(', ');
+}
+
+//         Initialization functions: called once per disaster.
```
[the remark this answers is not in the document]
janakdr: I think we can assume everything is scoped to page load. Our entire site is client-side, no?

docs/import/manage_disaster_flexible.js
janakdr: Because I think it's clearer to have the building-count column in the building-count source section, not with the poverty section. Do you think it would be better up with the poverty section? There would still have to be special-casing around its visibility, so it's a UX question, not code simplicity, I think.

docs/import/manage_disaster_flexible.js
janakdr: "selection no longer current" was my attempt at the description. Expanded a bit. Better?

docs/import/manage_disaster_flexible.js
janakdr: Rephrased comment. Any better?

docs/import/manage_disaster_flexible.js
janakdr: Good call.

janakdr:
Now with tons of tests. Tests caught a bug in my original approach: it required all pending operations to complete before we did validation, even ones that were no longer relevant, so that if one asset's pending operation never completed, but the user switched away from that asset, we'd still never do validation. Had to make the "pending" logic more complex, but hopefully well encapsulated. The "pending" operations of manage_disaster_flexible and manage_disaster_base (that's just damage) could use a new look. No more finally blocks, so that's nice maybe!

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+ * current disaster is allowed to do that.
+ *
+ * Similarly, if an asset value changes, we'll immediately kick off a query for
+ * the new values' columns. If the old value column query had not completed, we
+ * don't increment the pending count.
+ *
+ * There should be one `PendingChecker` for each operation that must finish
+ * before validation can run, but which can be rendered irrelevant by a user
+ * action. Thus, each select whose value cascades (poverty, geography, damage,
+ * buildings) must have a `PendingChecker`. Similarly initialization as a whole
```
juliexxia: initialization of what?

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+ * has a `PendingChecker`. This keeps our count of pending operations limited to
+ * the types of operations there can be, which is correct: the user can never be
+ * waiting on more than one operation of a given type.
```
juliexxia: type here being like geography asset collection right?

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+ * complete in the background and be ready, even if the user is switching other
+ * options back and forth.
+ */
+class PendingChecker {
+  pending = PendingState.NOT_PENDING;
+
+  /** @return {boolean} False if this checker was already pending */
+  maybeStartPending() {
+    if (!isFlexible()) {
+      return true;
```
juliexxia: Why early return if state based?

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+    }
+    switch (this.pending) {
+      case PendingState.NOT_PENDING:
+        this.pending = PendingState.PENDING;
+        startPendingOperation();
+        return true;
+      case PendingState.PENDING:
+        return false;
+      case PendingState.PENDING_BUT_INVISIBLE:
+        showError('Error validating inputs: please reload page');
```
juliexxia: Why does this throw an error?

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
 /**
  * Functions for dealing with a "flexible" (non-state-based) disaster.
  *
+ * ## Structure of a flexible disaster
+ *
+ * A flexible disaster takes more general input assets than a state-based one.
+ * We expect the following (analogs to the state-based case are in parentheses):
+ *
+ * 1. A poverty asset, which must have the following columns:
+ *      a. Poverty rate ('SNAP PERCENTAGE');
```
juliexxia: nit maybe prepend these examples with e.g. to make it clear it doesn't need to be those exact cols

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+
+ *    If the poverty asset does not have geometries, then we also need:
+ * 2. A geography asset, which must have geometries, and the following column
+ *    to join to poverty:
+ *      a. District identifier ('GEOID');
+ *
+ *    Before damage is present, that is all. With damage, we also need:
+ * 3. A building-count source to be specified. Either 'buildings', 'poverty', or
+ *    'damage'.
+ * 4. If the source is 'buildings', we need an optional buildings asset. If it
```
juliexxia: "we need an optional" is slightly confusing

docs/import/manage_disaster_flexible.js
```diff
@@ -1,22 +1,154 @@
...
+ *    has geometries, it is assumed to have building footprints (like the
+ *    Microsoft building footprints) and is intersected with district
```
juliexxia: might be nice to link to those footprints.

maybe rephrase to make it clear we do the intersecting for them, not that they already need to be intersected (not even sure what that would mean but)

docs/import/manage_disaster_flexible.js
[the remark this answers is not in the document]
juliexxia: I guess I think of creating a disaster as initializing a disaster

docs/import/manage_disaster_flexible.js
```diff
@@ -380,63 +524,74 @@ async function initializeFlexibleDisaster(assetData) {
...
-    finishPending();
+  initializeChecker.maybeStartPending();
+  const [result] = await Promise.all([
+    initializeGeography(),
+    initializePoverty(),
+    initializeBuildings(),
+  ]);
+  if (result) {
+    // Just check one value to see if it succeeded: others must be the same.
+    initializeChecker.finishPending();
```
juliexxia: Is the finishPending what's checking the value?

docs/import/manage_disaster_flexible.js
juliexxia: gotcha, I meant why can't it just go with the rest of the building stuff in the same div but this is fine

docs/import/manage_disaster_flexible.js
```diff
@@ -380,63 +524,74 @@ async function initializeFlexibleDisaster(assetData) {
...
   }
   // Set actual geography visibility.
   await showGeographyDivBasedOnPoverty(select.val());
-  const propertyNames = await setInitialColumnValues('poverty');
-  if (propertyNames) {
-    // Do special column setup, if disaster unchanged.
-    setOptionsForPovertyBuildings(propertyNames);
-  }
+  setInitialColumnValues('poverty').then((propertyNames) => {
+    if (propertyNames) {
```
juliexxia: why inside a then block now?, and sorry where are we checking disaster hasn't changed?

docs/import/manage_disaster_flexible.js
juliexxia: Yes, thanks!

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -17,613 +21,1431 @@ const scoreBoundsCoordinates = [
...
-    const buildingsPropertyPath =
-        stateBasedScoreAssetTypes[4].propertyPath.concat(['NY']);
-
-
-    checkSelectBorder(buildingsPropertyPath, 'rgb(0, 128, 0)');
-    checkHoverText(buildingsPropertyPath, 'No expected columns');
-    setFirstSelectInScoreRowTo(4, 'None');
-    checkSelectBorder(buildingsPropertyPath, 'rgb(255, 255, 255)');
-    checkHoverText(buildingsPropertyPath, '');
+  it('makes geography unneeded while pending', () => {
```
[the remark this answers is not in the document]
janakdr: Rephrased, better?

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -17,613 +21,1431 @@ const scoreBoundsCoordinates = [
...
+      'b',
+    ]));
+    propertyNamesStub.withArgs('asset2').returns(
+        new Promise((resolve) => resolveFunction = resolve));
+    cy.wrap(enableWhenFirestoreReady(
+        new Map([[getDisaster(), createDisasterData(null)]])));
+    getSelectFromPropertyPath(componentsData.poverty.path)
+        .select('noGeoAsset')
+        .blur();
+    cy.get('#kickoff-button').should('not.have.text', 'Pending...');
```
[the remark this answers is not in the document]
janakdr: Added comments to the first test of each pair. Better?

docs/import/manage_disaster.js
```diff
@@ -85,13 +83,15 @@ async function onSetDisaster() {
   const flexible = isFlexible(currentData);
```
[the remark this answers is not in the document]
janakdr: I think it's used twice, forgot to use the variable below, fixed.

docs/import/manage_disaster_flexible.js
janakdr: Added a word

docs/import/manage_disaster_flexible.js
janakdr: yes, added a few words

docs/import/manage_disaster_flexible.js
janakdr: Because this whole file is only for flexible disasters

docs/import/manage_disaster_flexible.js
janakdr: Added a developer error message. Better?

docs/import/manage_disaster_flexible.js
janakdr: Stale language, fixed.

docs/import/manage_disaster_flexible.js
janakdr: Seems doubtful someone reading this is interested in going there: can't actually check them out without a lot of preprocessing. Added breadcrumb to Data Source Guide. Rephrased intersection.

docs/import/manage_disaster_flexible.js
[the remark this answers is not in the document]
janakdr: If there's a word you prefer over initialize, I'm happy to use it. Added some text here.

docs/import/manage_disaster_flexible.js
janakdr: No, moved line up to hopefully make clearer.

docs/import/manage_disaster_flexible.js
janakdr: It can't go with the building div because this column is only relevant when poverty asset is used for buildings, and building div is hidden then. Not sure if that's what you were suggesting?

docs/import/manage_disaster_flexible.js
janakdr: Rephrased a bit, and fixed called functions' docs, which were stale about the fact that disaster change also triggers null return.

It's in a then block now because we're no longer waiting for it to finish before returning. The initialize* functions don't wait for column population, just for the asset pickers' population.

docs/import/manage_disaster_flexible.js
juliexxia: maybe rename to geographyResult

cypress/integration/unit_tests/manage_disaster_set_score_parameters_test.js
```diff
@@ -17,613 +21,1431 @@ const scoreBoundsCoordinates = [
...
+     featureCollectionStub.withArgs('asset1').returns(asset1);
+     featureCollectionStub.withArgs('asset2').returns(asset2);
+
+     // Set properties up so that poverty asset is found, income asset is not
+     // found, and svi is never set.
+     const currentData = createDefaultStateBasedFirestoreData();
+     currentData.assetData.stateBasedData.snapData.paths.NY = 'found-asset';
+     currentData.assetData.stateBasedData.incomeAssetPaths.NY = 'missing-asset';
+     currentData.assetData.damageAssetPath = 'missing-asset';
+     currentData.assetData.noDamageKey = 'a-key';
```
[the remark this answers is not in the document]
janakdr: Done.

PR #420, opened 2020-01-14
Update help links to point to GD copies and fix headers

ruthtalbot:


janakdr:
So we're definitely doing this rather than a wiki?

PR #424, opened 2020-01-15
Always use rgb values for colors 

juliexxia:
Greens weren't matching up before. I also am not clear how createCheckboxForUserFeatures was working before, so updated. 

Now that layer green is very similar to the green of user features, make user features a nice cyan ('#00FFFF')

image

### DIFF
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -148,7 +148,7 @@ function getLinearGradient(colorFunction) {
 /**
  * Converts a color to a css-recognizable rgb string.
  * @param {string} color one of the keys of our colorMap or a hex string
- * for special cases like the user features color or the score color.
+ *     for special cases like the user features color or the score color.
  * @return {string}
  */
 function colorToRgbString(color) {
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -125,7 +125,8 @@ function getLinearGradient(colorFunction) {
   let gradientString = 'linear-gradient(to right';
   switch (colorFunction.currentStyle) {
     case ColorStyle.CONTINUOUS:
-      gradientString += ', rgb(255,255,255), ' + colorToRgbString(colorFunction['color']);
+      gradientString +=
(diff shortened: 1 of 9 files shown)


docs/firebase_layers.js
```diff
@@ -123,21 +124,36 @@ function getLinearGradient(colorFunction) {
...
 }
+
+/**
+ * Converts a color to a css-recognizable rgb string.
+ * @param {string} color
+ * @return {string}
+ */
+function colorToRgbString(color) {
+  const rgbArray = colorMap.get(color);
+  if (!rgbArray) {
```
janakdr: this is an escape hatch because a lot of consumers can handle a color string?

docs/run.js
```diff
@@ -272,7 +276,7 @@ function addLayers(map, firebaseLayers) {
     }
     createNewCheckboxForLayer(properties, sidebarDiv, map);
   }
-  createCheckboxForUserFeatures(sidebarDiv);
+  createCheckboxForUserFeatures(sidebarDiv, map);
```
janakdr: why this change? I don't think it takes the map as an argument?

docs/firebase_layers.js
janakdr: Could you instead/in addition add documentation to the `@param` tag for color that says what it's expected to be?

juliexxia: Done

docs/firebase_layers.js
```diff
@@ -123,21 +124,37 @@ function getLinearGradient(colorFunction) {
...
+      gradientString += ', ' + rgb + ', ' + rgb;
       break;
   }
   return gradientString + ')';
 }
+
+/**
+ * Converts a color to a css-recognizable rgb string.
+ * @param {string} color one of the keys of our colorMap or a hex string
+ * for special cases like the user features color or the score color.
```
janakdr: nit: I think this should be indented 4 spaces.

Issue #406, opened 2020-01-09
Default to most recent disaster on first visit to page (not harvey)


juliexxia:

janakdr:
Anything left here?

PR #432, opened 2020-01-16
factor out default disaster to standalone file for easy access

juliexxia:
@ruthtalbot not sure where documentation should go for this. I documented lightly in file but I figure more extensive docs will have to live somewhere else

### DIFF
--- a/docs/default_disaster.js
+++ b/docs/default_disaster.js
@@ -1,6 +1,6 @@
 /**
  * The default disaster that will show up the very first time a user visits
- * the web app. In order to update which disaster users land on the first time,
+ * the map. In order to update which disaster users land on the first time,
  * update the string below and commit to the github repository. The string
  * should always be in <year>-<name> format (with the name being all lowercase).
  *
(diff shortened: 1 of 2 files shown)


docs/resources.js
```diff
@@ -7,11 +8,11 @@ export {getBackupScoreAssetPath, getDisaster, getScoreAssetPath};
  * @return {string} current disaster
  */
 function getDisaster() {
-  if (localStorage.getItem('disaster')) {
+  if (localSgtorage.getItem('disaster')) {
```
janakdr: Typo

docs/default_disaster.js
```diff
@@ -0,0 +1,14 @@
+export {
+  DEFAULT_DISASTER
```
janakdr: Can we one-line this?

Also, could violate convention and put at end of file so line user has to edit is closer to top?

juliexxia: I thought about that too. Went for it!

janakdr:
#406

PR #433, opened 2020-01-16
Factor out code references to GD to a standalone file

juliexxia:
Also minimal docs here

### DIFF
--- a/docs/authenticate.js
+++ b/docs/authenticate.js
@@ -3,9 +3,11 @@ import {eeLegacyPathPrefix, eeLegacyPrefix} from './ee_paths.js';
 import {showError} from './error.js';
 import {listEeAssets} from './import/ee_utils.js';
 import {earthEngineTestTokenCookieName, firebaseTestTokenPropertyName, getValueFromLocalStorage, inProduction} from './in_test_util.js';
+import {CONTACT, OWNER} from './owner.js';
 import {getBackupScoreAssetPath, getDisaster, getScoreAssetPath} from './resources.js';
 import {SettablePromise} from './settable_promise.js';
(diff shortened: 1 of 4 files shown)


docs/authenticate.js
```diff
@@ -179,12 +178,12 @@ class Authenticator {
...
             const message = 'Refresh token error: ' + response.status;
             console.error(message, response);
+            const contact = CONTACT ? CONTACT : OWNER;
             // TODO(#395): Find GD contact to list here.
             alert(
                 'Error contacting server for access without EarthEngine ' +
                 'whitelisting. Please reload page and log in with an ' +
-                'EarthEngine-whitelisted account or contact ' +
-                'technology.manager@givedirectly.org ' +
+                'EarthEngine-whitelisted account or contact ' + contact + ' ' +
```
janakdr: need to ALL-CAPS contact?

token_server/token_server.js
```diff
@@ -4,6 +4,7 @@ import {createServer} from 'http';
 import parseBody from 'urlencoded-body-parser';
 import {storeGoogleCredentials} from './aws_get_credentials.js';
 import {generateEarthEngineToken} from './ee_token_creator.js';
+import {TOKEN_SERVER_ORIGINS} from '../docs/owner.js';
```
janakdr: I think deploying token server doesn't work with local files outside this directory, sadly :(

docs/authenticate.js
juliexxia: I variablized it outside

token_server/token_server.js
juliexxia: Sad, moved to a file in this dir

PR #434, opened 2020-01-16
Help link updates [attempt #2]

ruthtalbot:


janakdr:
Hmm, I'm not really familiar with that code. May have never worked. If it did work, we should remove the hard-coded link in navbar.html. @RamV13 ?

RamV13:
As of #291, the `getHelpUrl` made the main help link in the navbar page specific by updating `help-a` in `loadNavbar` within `navbar.js`. Something seems to have broken this since then.

RamV13:
Worth noting that this is broken even at head so probably not because of this change.

RamV13:
It's because there is a one `a` with a `help-a` class and another with a `help-a` id.

RamV13:
Ah just missed your comment cool :)

PR #437, opened 2020-01-17
Update EarthEngine version to latest

janakdr:
EarthEngine now requests additional permissions, which Google.com corp accounts are not allowed to grant to third-party apps. As a workaround, use the token server if we get an error indicating that authentication failed due to a permissions error like this.

Images are loaded a bit differently now (`ee.MapLayerOverlay` is apparently deprecated), so had to modify layer_util.js. Is cleaner now.

Also noticed that after a minor version node upgrade (12.15-16), our tests started failing, so pinning at node v12.15. I think 12.16.1 should fix, but not important. More important is emailing test failures to GD when they happen.


docs/import/manage_disaster.html
```diff
@@ -17,10 +17,11 @@
   <link href="../external_libs/jquery-ui-1.12.1.min.css" rel="stylesheet">
 
   <!-- Load debug Earth Engine JavaScript client library. -->
-  <!-- TODO(janakr): Switch back to compiled version in production. -->
+  <!-- TODO(janakr): Switch back to compiled version in production. Compiled
```
juliexxia: Still relevant TODO? When do you think this will be todone?

janakdr: See internal thread. They say this or next week?

janakdr:
Updated to latest version of EE. Now actually works! Main change from previous version is in layer_util.js.

docs/layer_util.js
```diff
@@ -268,14 +267,26 @@ function showOverlayLayer(overlay, index, map) {
 function resolveOnEeTilesFinished(layerDisplayData, resolve) {
   if (layerDisplayData.tileCallbackId) {
     layerDisplayData.overlay.removeTileCallback(
-        layerDisplayData.tileCallbackId);
+        layerDisplayData.tileCallbackId[0]);
```
juliexxia: This method seems to be more than a wrapper now - update doc to reflect?

cypress/integration/unit_tests/checkbox_test.js
```diff
@@ -305,8 +305,10 @@ describe('Unit test for toggleLayerOn', () => {
         })
         .then(() => {
           expect(loadingFinishedStub).to.be.calledOnce;
+          expect(map.overlayMapTypes).to.have.length(4);
           overlay = map.overlayMapTypes.getAt(3);
           expect(overlay).is.not.null;
+          expect(overlay).to.not.be.undefined;
```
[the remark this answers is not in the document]
janakdr: Done.

docs/layer_util.js
janakdr: Done, and fixed stale reference in doc. Thanks!

docs/import/ee_utils.js
```diff
@@ -19,6 +19,7 @@ async function listEeAssets(path) {
     let assets;
     const listResult =
         await ee.data.listAssets(path, {page_token: nextPageToken}, () => {});
+    console.log(listResult.assets);
```
juliexxia: rm

janakdr: Oops, done!

PR #441, opened 2020-01-17
Support multiple Census columns, and other clean-ups

### LINKED ISSUE #430: Support data.census.gov column headers/geoid format
- [x] parse the longform geoid (with some prefix and 'US' in it before the geoid we've been using)
- [x] allows for census.data.gov columns

janakdr:
* We now support both the old and new Census column names. If neither is found for an asset, we show an error in the UI, telling user the possibilities there can be.
* As part of this, get rid of the (never exposed) flexibility in Census column names: stop retrieving them from Firestore. A truly flexible disaster can be created if needed.
* Change 'GEOID' column used internally to have a likely-unique name to avoid clashing with columns naturally found in user data.
* Change 'Poverty' row to be called 'SNAP', since it's always SNAP. (Feedback from GD meeting.)
* Prepend disaster name to EE export task so it's easier to identify them

The multiple-column feature is a little peculiar in that it does the fallback on a feature-by-feature and column-by-column basis. Thus, if we had some weird Frankenstein FeatureCollection where features came from different Census datasets, it would still work. But not worth trying to validate the whole FeatureCollection: we just continue doing the same validation on the first Feature's columns (which have to all belong to one schema or the other) and during score asset creation we proceed blindly.

The migration story is a bit tricky for this, because our current assets have a GEOID column. My plan is to create new assets with the new GEOID column name and swap them into place as I submit this PR.

New disaster (first image) looks reasonably like old one (second image) without damage taken into account.

image

image

Closes #430


docs/import/create_disaster_lib.js
```diff
@@ -2,18 +2,12 @@ export {createDisasterData};
 export {
   BUILDING_COUNT_KEY,
   BuildingSource,
-  incomeKey,
-  snapKey,
-  sviKey,
-  totalKey,
+
```
juliexxia: nit: multiple \n

docs/import/create_score_asset.js
```diff
@@ -220,13 +234,33 @@ function stringifyCollection(
...
+ */
+function setGeoid(censusTable) {
+  return ee.FeatureCollection(censusTable).map((f) => {
+    // Starts with 150000000US, indicating block groups.
+    // https://www.census.gov/programs-surveys/geography/guidance/geo-identifiers.html
+    const geoid =
+        ee.String(getKeyOrBackup(f, censusGeoidKey, backupCensusGeoidKey));
+    return f.set(
+        geoidTag,
+        // If too short, null out: can happen with header rows.
```
juliexxia: Sorry, not totally understanding the null out scenario (is it expected?) And how does this handle the old version?

docs/import/manage_disaster_base.js
```diff
@@ -392,11 +395,15 @@ async function verifyAsset(propertyPath, expectedColumns) {
...
     updateColorAndHover(select, 'white', '');
     return [];
-  } else if (!expectedColumns || expectedColumns.length === 0) {
-    updateColorAndHover(select, 'yellow', 'Checking...');
+  } else {
+    const checkingColumns = expectedColumns && expectedColumns.length > 0;
+    if (checkingColumns) {
+      updateColorAndHover(select, 'yellow', 'Checking columns...');
+    } else {
+      updateColorAndHover(select, 'yellow', 'Checking...');
```
juliexxia: What checking do we need to do in this case? Why do we not go straight to green?

docs/property_names.js
```diff
@@ -8,7 +8,8 @@ export {
 };
 
 const damageTag = 'DAMAGE PERCENTAGE';
-const geoidTag = 'GEOID';
+// Unique name to avoid clashing with user-given columns.
```
juliexxia: Can you add a note that this is never actually exposed to the end user

docs/import/create_disaster_lib.js
janakdr: weird!

docs/import/create_score_asset.js
janakdr: Expanded comment a bit. Better? Old version also has this geoid.

docs/import/manage_disaster_base.js
janakdr: Whether or not it exists.

This is just a refactoring of the existing code because we're getting names unconditionally now, but users were already seeing this.

docs/property_names.js
janakdr: Done.

PR #449, opened 2020-02-04
Fix color rendering

janakdr:
Don't accidentally modify the color arrays.

I wonder if the latest deck.gl that we started using a few weeks before fellowship ended got stricter? It looks like this bug has been here since this code was introduced in #186.

### DIFF
--- a/docs/firebase_layers.js
+++ b/docs/firebase_layers.js
@@ -39,8 +39,7 @@ function createStyleFunction(colorFunctionProperties) {
   const field = colorFunctionProperties['field'];
   switch (colorFunctionProperties.currentStyle) {
     case ColorStyle.SINGLE:
-      // Drop opacity for single-color.
-      const color = colorMap.get(colorFunctionProperties.color).slice(0, 3);
+      const color = colorMap.get(colorFunctionProperties.color);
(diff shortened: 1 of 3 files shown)


docs/firebase_layers.js
```diff
@@ -36,7 +39,8 @@ function createStyleFunction(colorFunctionProperties) {
   const field = colorFunctionProperties['field'];
   switch (colorFunctionProperties.currentStyle) {
     case ColorStyle.SINGLE:
-      const color = colorMap.get(colorFunctionProperties.color);
+      // Drop opacity for single-color.
```
juliexxia: why?

janakdr: Just preserving behavior: we didn't specify opacity before, which made it solid.

juliexxia: If it's not too much work for you, I'd rather update single-color to use opacity. Will let us drop this special check and standardize across coloring schemes. If that's going to require a more extensive PR though happy to approve as is.

janakdr: Just changing the special-casing is easy (until Travis tells me otherwise), done.

PR #450, opened 2020-02-04
Close hyperlink on manage_disaster page /shamecube

janakdr:


juliexxia:
Haha. Is this why it was refreshing repeatedly? Very strange front end reaction if so.

janakdr:
I think the unclosed link made the whole table one big hyperlink, but with a blank target

PR #452, opened 2020-05-11
Properly wait for response.json() promise

janakdr:
Broken by https://github.com/givedirectly/Google-Partnership/commit/72dd1775cf549bec616f39a6467d73d3aaa2f42d#diff-10d6fe37a932cbb977d8ff39d67db610L203: we used to chain promises. When I switched to async style as part of EE upgrade, forgot to add "await" here.

### DIFF
--- a/token_server/token_server.js
+++ b/token_server/token_server.js
@@ -104,7 +104,7 @@ function generateTokenPeriodically() {
   createServer(async (req, res) => {
     const origin = req.headers['origin'];
     if (!allowedOrigins.has(origin)) {
-      console.warn("Bad origin", origin);
+      console.warn('Bad origin', origin);
       fail(res);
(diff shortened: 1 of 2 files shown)


token_server/token_server.js
```diff
@@ -104,6 +104,7 @@ function generateTokenPeriodically() {
   createServer(async (req, res) => {
     const origin = req.headers['origin'];
     if (!allowedOrigins.has(origin)) {
+      console.warn('Bad origin', origin);
```
juliexxia, line 107: Did you mean to leave this in? Fine by me if so

janakdr: yeah, thought it would be helpful for future issues, even though it wasn't the cause here.

PR #454, opened 2020-08-26
Lower max for toList

janakdr:
Seems to be tickling an EarthEngine bug.

### DIFF
--- a/docs/ee_promise_cache.js
+++ b/docs/ee_promise_cache.js
@@ -9,7 +9,7 @@ export {
 export {clearPromiseCacheForTesting};
 
 // 6M objects in a FeatureCollection: seems to be an EE bug that larger numbers
-// trip it up.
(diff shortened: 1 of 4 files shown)


docs/ee_promise_cache.js
```diff
@@ -8,8 +8,9 @@ export {
 // For testing only.
 export {clearPromiseCacheForTesting};
 
-// 250M objects in a FeatureCollection ought to be enough for anyone.
-const maxNumFeaturesExpected = 250000000;
+// 6M objects in a FeatureCollection: seems to be an EE bug that larger numbers
```
juliexxia, line 11: Can you add that this was og 250000000 so reader has an idea what larger numbers means here

janakdr: Done.

PR #459, opened 2020-10-25
upgrade earthengine client library to 0.1.232

janakdr:
There were some changes that didn't seem related to this upgrade. They may have come from my running tests on my Mac on the Electron browser: we might have let coverage slip on that. Folded them in here.


janakdr:
Jumped the gun a bit on sending the PR out: didn't check that I could create a new score asset. Currently running that, will update here when that completes.

janakdr:
Took longer than I expected (almost an hour for my toy disaster) but worked ok, I think.

juliexxia (approved):
Thanks!

PR #464, opened 2022-10-11
Drop "404" error pages when fetching image tiles

janakdr:
This avoids ugly "broken image" icons like
image

Do a belt-and-suspenders fix to also hide the image if it is in error. That seems less brittle to me than examining the type of the downloaded blob, but unfortunately it causes flickering since the image renders broken and only then gets hidden.

Also fix a minor bug (probably introduced in https://github.com/givedirectly/Google-Partnership/pull/437) that caused the loading indicator overlay to be prematurely removed and then never added back.

Not adding tests in the interests of expediency and also I think tests might be broken right now.


janakdr:
Will still take a post facto review :)

PR #466, opened 2024-06-06
Fix tests by upgrading Cypress to 11.2.0

janakdr:
Cypress needed to be upgraded because the Google Maps Javascript library was too advanced for the version of Javascript run by Cypress, and the normal trick of using Babel transpilation to make it compatible didn't work with the dynamic loading of dependencies the Google Maps javascript library was doing.

The main change needed to upgrade Cypress was moving to `cypress.config.js` as the config file (instead of `index.js` and `main.js`), which also allowed us to use Javascript modules, rather than the terrible hacks we had going before.

With Cypress upgraded, and since the tests hadn't run for years, I had to make a few changes. In descending order of invasiveness:

(1) (Prod change!) The JSON tile format can now accept a list of JSON specifiers, since that's what NOAA provides for recent disasters, rather than a single JSON map with potentially multiple tiles. Production change in `layer_util.js` with test change in `checkbox_test.js`.
(2) In `polygon_draw_test.js`, it seems like the test was wrong for quite a while: a [comment about the polygon intersecting two features](https://github.com/givedirectly/Google-Partnership/commit/d7735a5701e27dd7988cf9d1b686c7ebd94ec551#diff-c8240e25a606de5dc8243134810f8b755e72935e9e68715a57a2e3f2e5bc71fd) (click on "Load diff") was blindly copied over from the integration test while not being true anymore. I think we were getting lucky maybe due to rounding? I've changed the test parameters to be clearer.
(3) Type checking appears to have gotten stricter, either because of Javascript versions or library changes (although some of the type checking is happening in EarthEngine, which should have a frozen library). Because of that, `click_feature_test.js` as well as `polygon_draw_test.js` needed to be modified to get sneakier around type faking.
(4) We were incorrectly using `cy.get()` (which always searches from the root of the document) instead of `cy.find()` (which searches from the element it's chained off of). This was making some tests pass incorrectly, and making others fail flakily. Changes in `highlight_features_test.js`, `manage_disaster_set_score_parameters_test.js`.
(5) Some EarthEngine error messages appear to have changed. I loosened the assertions in `checkbox_test.js`.
(6) Upgrading Cypress resolved a TODO in `run_test.js`.
(7) Deleted empty `navbar_test.js`.

Other changes: I upgraded most packages as far as I could without getting errors: package.json now contains `<=` for a version if I was unable to upgrade it past that version. Firebase beyond this version causes module-style errors in Cypress. Cypress beyond this version causes weird errors in some tests. eslint beyond this version uses a new config file format.

`cypress.config.js` should be branched from `cypress/plugins/main.js` but git wasn't smart enough to know that (or I wasn't smart enough to tell git that).


janakdr:
cc @juliexxia, I couldn't add you as a reviewer for whatever reason

crystal-li (approved):
Tbh didn't look super closely but overall checked that it matches what you outlined in the comments! Thanks Janak!

janakdr:
@juliexxia I'm submitting, but please feel free to review. If you see any issues, happy to fix in a follow-up.

PR #467, opened 2024-06-10
Create cypress workflow

janakdr:
Add a Cypress workflow based on https://docs.cypress.io/guides/continuous-integration/github-actions. Also add in our linter.

Running this and debugging it revealed some issues, also fixed here:

* Fix missing googleapis dependency (it's only needed for the token server, but the token server library is loaded by Cypress).
* Use proper launchOptions object in browser options modification (https://docs.cypress.io/guides/references/migration-guide#Plugin-Event-beforebrowserlaunch).
* Remove stray console logging of config during Cypress startup.
* Change documentation to remove references to Travis and include instructions for setting this up.
* Reformat `cypress.config.js`.


janakdr:
cc @juliexxia

crystal-li (approved):
Ach my bad - this dropped off my radar last week.

Appreciate your thorough PR descriptions - makes it much easier for me to review. Also thanks for going through and updating docs / any references to Travis 🙏🏼

PR #468, opened 2024-10-17
Augment error message when there are non-matching columns.

janakdr:
This makes it easier to understand why an asset has an error. In real life, an asset had non-ascii characters in the column name, so add a specific check for that, because when the character is non-printable, it's impossible to tell with the naked eye why the name is wrong.

### DIFF
--- a/docs/import/manage_disaster_base.js
+++ b/docs/import/manage_disaster_base.js
@@ -69,6 +69,8 @@ const DAMAGE_PROPERTY_PATH = Object.freeze(['damageAssetPath']);
 
 const damageAssetChecker = new PendingChecker();
 
+const asciiRegex = /^[ -~]+$/;
+
 /**
  * Does all initialization for damage asset and related fields. Creates damage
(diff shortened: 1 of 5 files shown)


docs/import/manage_disaster_base.js
```diff
@@ -419,34 +423,63 @@ async function verifyAsset(propertyPath, expectedColumns) {
...
+        return left[0].length < right[0].length ? -1 : 1;
+      });
+      let errStr = 'Error! asset is missing columns for all ' +
+          expectedColumns.length + ' possibilities for valid columns:\n';
+      for (const [missing, ind] of missingColumns) {
+        errStr +=
+            '* [' + expectedColumns[ind] + '] (missing ' + missing + ')\n';
+      }
+      if (badColumns.length > 0) {
+        errStr += 'Columns with non-ascii characters: ' + badColumns;
```
crystal-li, line 462: 🙏🏼

crystal-li (approved):
Thank you!

PR #469, opened 2024-10-17
Use damage asset's building count if it's greater than building asset's

janakdr:
If the damage asset has a more granular picture of the buildings in a block group and it also includes undamaged buildings, use the total number of buildings in the group according to the damage asset as the source of truth. This avoids weird things like >100% damage.

This could be accomplished with another knob on the "manage_disaster" page, similar  to the flexible disaster page, where the user can choose between getting the building count from a separate asset, the poverty asset, or the damage asset. However, it seems safe enough to do this behind the scenes. The two risks are: if the damage asset has lots of points that *aren't* actually buildings but are listed as undamaged (like, for whatever reason every fire hydrant is marked as an undamaged point).  Then the number of buildings according to the damage asset will be an overestimate, even if the damaged building count is accurate. The other risk is if the damage asset has all the buildings in some areas, but not in others. Then we can't really compare across areas. Both of these seem unlikely, and they can be fixed by cleaning up the damage asset (although detecting it might be tough).

Tests to come in a follow-up, since this is somewhat urgent!


janakdr:
cc @juliexxia

PR #470, opened 2024-10-18
Actually store the new building count

janakdr:
Forgot that EarthEngine "set" is really just a transformation.

Tested this in the EE code editor, but this shows I should have written a test for the last PR...


crystal-li (approved):
Oops my bad too for not looking carefully here! Ty for fix!

PR #473, opened 2025-04-14
Allow zero-score tracts to be displayed in the table

### LINKED ISSUE #472: User wants to be able to see tract info where Score = 0
Repro:
1. Log in as gd-earthengine-user@givedirectly.org
2. Go to any map area and set the poverty threshold to 0.
3. Find a tract where "score = 0".
4. You'll find that all tracts where score = 0 do not show up in the table below, and are not exported as part of the CSV when you hit "'Download"  (in the screenshot below, I sorted "score" in ascending order and only non-zero values are included).

Use case: For certain disasters, we want to be able see data (e.g. SVI) for all tracts in an area, even if the poverty score is 0.

Image

janakdr:
The table was using zero score to indicate that a block group was filtered out. That's normally true (we set the score to 0 for any filtered-out block group), but if the filtering options are set to include everything, some block groups that have a genuine zero score won't be included.

Also sneak in a fix to score_bounds_map_test since Google Maps changed the "Zoom in" button to not be visible by default.

Fixes #472

### DIFF
--- a/docs/draw_table.js
+++ b/docs/draw_table.js
@@ -1,8 +1,10 @@
 import {showError} from './error.js';
 import {highlightFeatures} from './highlight_features.js';
-import {displayedTag} from './property_names.js';
+import {displayedTag, geoidTag} from './property_names.js';
 
 export {drawTable};
+
+const hiddenColumns = new Set(geoidTag, displayedTag);
 
 /**
  * Displays a ranked table of the given features that have non-zero score. Sets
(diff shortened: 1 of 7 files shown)


docs/draw_table.js
```diff
@@ -57,8 +57,8 @@ async function drawTable(scoredFeaturesAndColumns, map) {
 function renderTable(list, features, map, selectorReceiver) {
   const data = google.visualization.arrayToDataTable(list, false);
   const dataView = new google.visualization.DataView(data);
-  // don't display geoid
-  dataView.hideColumns([0]);
+  // don't display geoid or displayedTag.
+  dataView.hideColumns([0, 1]);
```
crystal-li: nit: Is there a way to reference these by column name instead of just index? (not a big deal because we don't change this code that much but in case there is an easy way)

crystal-li (approved):
thanks for fixing!

docs/draw_table.js
janakdr: Sure, done.

